PR #17, opened 2024-07-30
update all links that refer to the repo itself

rcreasi:


pgm (approved):
oooooooh, I didn't think think of looking for links to the old repo. Thanks for updating these @rcreasi !

naquib314 (approved):
Looks good to me :)

PR #20, opened 2024-08-02
Make select in DE2 editable after a selection has been made

rcreasi:
This modifies the behavior of all select components in DE2's configuration panel. Before it was not possible to edit a value once it was selected. Now the cursor is automatically placed at the end of the value and the user can use the arrow keys or backspace key to modify it.

How is this useful to users? Consider this scenario:
- I select gene ABCD1
- I decide I actually want to look at gene ABCD2
- I can now press backspace once and type a '2' in place of the '1'
modify

"Edit mode" doesn't take effect until you start typing. This is to preserve the old behavior in cases where there is only a short list of features to choose from (and you never have to type anything at all).
short-list

This also works with really long feature names that overflow the container.
long-name

My dream is to preserve the line wrapping even after you press a key. Unfortunately, an HTML `<input>` element refuses to wrap. I experimented with rendering a `<textarea>` instead but that entirely freaked out the `react-select` library. I'm going to punt on that for now and circle back to it later.

### DIFF
--- a/frontend/packages/@depmap/data-explorer-2/src/utils/extend-react-select.tsx
+++ b/frontend/packages/@depmap/data-explorer-2/src/utils/extend-react-select.tsx
@@ -56,8 +56,10 @@ const ConditionalTooltip = ({
 // - Shows a tooltip when the value has been truncated
 // - Automatically set `menuPlacement` to "top" when the dropdown is near the
 // bottom of the viewport.
+// - Allows a pre-selected value to be edited as text (simulating a vanilla
+//   <input> element)
 // - Uses a windowed menu list. This allows thousands of options to be
-// displayed performantly (powered by react-windowed-select).
+//   displayed performantly (powered by react-windowed-select).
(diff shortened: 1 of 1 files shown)


frontend/packages/@depmap/data-explorer-2/src/utils/extend-react-select.tsx
```diff
@@ -189,10 +193,87 @@ export default function extendReactSelect(
                 }),
                 ...props.styles,
               }}
+              onFocus={(e) => {
+                if (!value?.label) {
+                  return;
+                }
+
+                // HACK: Make it appear as if the cursor is at the end of the
```
pgm, line 201: wow, crazy... I certainly don't know of a better alternative so I'll go with, if it looks right, it is right. Interacting with it will teach us whether there are any additional edge cases that we need to handle or not.

I appreciate the comment explaining how this works.

pgm (approved):

PR #25, opened 2024-08-06
Update interface of DE2 Context endpoints 

snwessel:
**Part 1 of my refactoring efforts to support the DE2 "context" endpoints in breadbox** ([Asana](https://app.asana.com/0/1165651979405609/1207979348562841/f)). I'm trying to update the contract of these endpoints to be more in line with what we want long term (removing unused parameters, etc.)

_I would appreciate any feedback on this!_ My next step will be to implement this same interface in breadbox, after which, it will be more difficult to change. 

I had initially planned to combine these two endpoints because they seem very closely related (both loading labels for a given a context). After looking through the code a bit more, I convinced myself that they're sufficiently different to keep separate, but could benefit from updated contracts. 

### The endpoints before/after my changes:
* `/evaluate_context` 
    * **Before my changes:**
        * Parameters: A "context" object and a boolean `"summarize"`, which determined which response type was returned
        * Returned: _either_ a dictionary listing all labels and aliases OR a dictionary listing the number of matches and number of "candidates"
    * **After my changes:** (now `/context/labels` )
        * Parameters: a "context" object
        * Returns: a dictionary like: 
        ```
        {
            "labels": [... ],
            "aliases": [...], # I'd like to eventually remove this from the response
            "num_candidates": 25, # the total number of labels belonging to the dimension type
         }
         ```
        * Note: I would like to remove the "aliases" from this response but they're still being used in the context manager downloads (specifically for cell line labels). I'll make a follow up task to remove the references to this so we don't need to support the "aliases" in the breadbox response. 

* `/datasets_matching_context`
    * **Before my changes:**
        * Parameters: a "context" object and a boolean `include_matching_entities`, which was always set to True. 
        * Returned: a list of dictionaries like
        ```
            [
                {
                  "dataset_id"    : "Chronos_Combined"
                  "dataset_label" : "CRISPR (DepMap Internal 23Q4+Score, Chronos)"
                  "entity_labels" : ["SOX10"] # if include_matching_entities = True
                },
                ...
            ]
        ```
    * **After my changes:** (now `/context/datasets` )
         * Parameters: only a "context" object 
         * Returns the same thing (now always including entity_labels)

### DIFF
--- a/frontend/packages/portal-frontend/src/data-explorer-2/components/ContextManager/DownloadContextModal.tsx
+++ b/frontend/packages/portal-frontend/src/data-explorer-2/components/ContextManager/DownloadContextModal.tsx
@@ -43,9 +43,10 @@ function DownloadContextModal({
 
     fetchContext(contextHash)
       .then(fetchContextLabels)
-      .then(async (evaluated) => {
-        let labels = evaluated.labels;
-
+      .then(async (depmap_ids) => {
+        // For now, we assume that the downloads are always depmap_models
+        // And the actual labels need to be loaded with a separate request
+        let labels = [...depmap_ids];
         if (include === "display_name" || include === "both") {
           if (context_type !== "depmap_model") {
             throw new Error("only supports depmap_model");
@@ -67,7 +68,7 @@ function DownloadContextModal({
         if (format === "csv" && include === "both") {
           text = [
             "DepMap ID,cell line name",
-            ...labels.map((label, i) => `${evaluated.labels[i]},${label}`),
+            ...labels.map((label, i) => `${depmap_ids[i]},${label}`),
           ].join("\r\n");
 
           download = `${filename}.csv`;
--- a/frontend/packages/portal-frontend/src/data-explorer-2/components/ContextManager/DownloadContextModal.tsx
+++ b/frontend/packages/portal-frontend/src/data-explorer-2/components/ContextManager/DownloadContextModal.tsx
@@ -4,6 +4,7 @@ import {
   fetchContextLabels,
   fetchContext,
   getDimensionTypeLabel,
+  fetchMetadataColumn,
   pluralize,
 } from "@depmap/data-explorer-2";
 import ContextNameForm from "src/data-explorer-2/components/ContextBuilder/ContextNameForm";
@@ -15,12 +16,6 @@ interface Props {
   contextHash: string;
   onHide: () => void;
 }
-
-const bySliceId = (slice_id: string) => (obj: unknown) =>
-  obj !== null &&
-  typeof obj === "object" &&
-  "slice_id" in obj &&
-  obj.slice_id === slice_id;
 
 function DownloadContextModal({
   contextName,
@@ -48,17 +43,21 @@ function DownloadContextModal({
 
     fetchContext(contextHash)
       .then(fetchContextLabels)
-      .then((evaluated) => {
+      .then(async (evaluated) => {
         let labels = evaluated.labels;
 
         if (include === "display_name" || include === "both") {
           if (context_type !== "depmap_model") {
             throw new Error("only supports depmap_model");
           }
-
-          labels = evaluated.aliases.find(
-            bySliceId("slice/cell_line_display_name/all/label")
-          )!.values;
+          const sliceId = "slice/cell_line_display_name/all/label";
+          labels = await fetchMetadataColumn(sliceId).then(
+            (cell_line_labels) => {
+              return labels.map(
+                (depmap_id) => cell_line_labels.indexed_values[depmap_id]
+              );
+            }
+          );
         }
 
         const link = document.createElement("a");
--- a/frontend/packages/portal-frontend/src/data-explorer-2/components/ContextManager/DownloadContextModal.tsx
+++ b/frontend/packages/portal-frontend/src/data-explorer-2/components/ContextManager/DownloadContextModal.tsx
@@ -1,7 +1,7 @@
 import React, { useEffect, useRef, useState } from "react";
 import { Button, Modal, Radio } from "react-bootstrap";
 import {
-  evaluateContext,
+  fetchContextLabels,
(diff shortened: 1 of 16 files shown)


portal-backend/depmap/data_explorer_2/views.py
```diff
@@ -444,49 +441,56 @@ def unique_values_or_range():
...
-            labels.append(label)
+            labels_matching_context.append(label)
+    aliases = get_aliases_matching_labels(context_type, labels_matching_context)
 
-    aliases = get_aliases_matching_labels(context_type, labels)
+    # Aliases should be removed from this response once they're no longer needed by context manager.
+    # The breadbox implementation of this endpoint will not return aliases.
+    response = {
+        "labels": labels_matching_context,
+        "aliases": aliases,
```
snwessel, line 468: Fyi, here is the one place that these "aliases" are still used in the frontend: https://github.com/broadinstitute/depmap-portal/blob/f7aadc1df076b840b059a3e0900dca15da1d1ed1/frontend/packages/portal-frontend/src/data-explorer-2/components/ContextManager/DownloadContextModal.tsx#L59

@rcreasi I think I remember you saying that this is something that could be replaced with a call to a different endpoint - is that right? If so, I can make a follow-up task in Asana.

rcreasi:
I don't love that `summarize` has gone away 😢 

Each of these rules gets evaluated individually to show sub-totals.
Screenshot 2024-08-06 at 5 08 51 PM

What used to be a few hundred _bytes_ going over the wire...
before

Is now thousands of _kilobtyes_
after

pgm:
> I don't love that `summarize` has gone away 😢

I'm a big advocate for minimizing API surface areas, however, this is an instance where I think we wouldn't be painting ourself in the corner if we added a "summarize" endpoint because it's something that would trivially be computed from the other. From the perspective of long term commitment, having both doesn't feel like a bigger commitment to only having one.

(Note: I said "summarize endpoint" but the way this currently works is by passing a flag. If we are changing contracts, I'd encourage having summarize handled as a separate endpoint because I have a bias towards making APIs with consistent return types. That is to say, currently the contract is something like `evaluate_context(context: Context, summarize: bool) -> ( MembershipDetails | Summary )` and to make it easier to reason about/type check I prefer having two methods, each with its own return type such as: `evaluate_context(context: Context) -> MembershipDetails` and `evaluate_context_summary(context: Context) -> Summary`

rcreasi:
> (Note: I said "summarize endpoint" but the way this currently works is by passing a flag. If we are changing contracts, I'd encourage having summarize handled as a separate endpoint 

A separate endpoint is a fine idea. I did something similar on the frontend where I split it into two separate API methods.

Another reason I'm keen to keep the summary behavior around is that leaves the door open to future optimizations. If you know you only need a count you can consume less memory and optimize away certain logic branches.

The code that evaluates contexts is not very efficient. It does the job for now but I had always envisioned translating the rules into SQL queries or the like. Then it literally could be a `count()` operation.

pgm (approved):
I'm approving to record I don't have any objections to anything I see here -- at the same time I have to note I don't have a lot of knowledge about how this all works, so I can't contribute anything beyond a shallow read through.

I think @rcreasi has a fair point about it being nice to have some way to get a summary. 

Also, if the goal is to define a contract prior to implementing something in breadbox, I'd argue for removing `alias` and switching the names in the API to be more consistent with breadbox now rather than later. 

But realistically, no objection to this PR as an increment on the path to getting the same API in breadbox.

portal-backend/depmap/data_explorer_2/views.py
snwessel: Sort of, with the current implementation:
* For features, the actual labels are returned
* For samples, the given_ids are returned as `labels` (this is the part we were talking about earlier), and I believe that's why the "aliases" are used. 

If I remember correctly, the reason we do this is because "sample labels" historically, were not easily supported in the legacy backend. Luckily, I think we're at a point now where:
1. The portal-backend could return actual sample labels without too much effort
2. It will be easy for breadbox to return actual sample labels

Maybe I'll add a topic to our Breadbox meeting to talk about this more. At some point, soon we should make the switch to returning actual sample labels - and perhaps that needs to happen before these aliases can be replaced.

rcreasi: Yes, there's a clear path to making the frontend no longer rely on aliases. There's already this `/get_metadata` endpoint.

https://github.com/broadinstitute/depmap-portal/blob/f7aadc1df076b840b059a3e0900dca15da1d1ed1/portal-backend/depmap/data_explorer_2/views.py#L153

You give it a slice ID and get back indexed values. That's what powers the "color by property" functionality. We could just leverage that to get a mapping from DepMap ID to cell line name by using it with the slice ID `slice/cell_line_display_name/all/label`.

snwessel:
By the way, I updated a few things and then re-tested that Context Manager is still working fine (including the downloads):
* I added the `/context/summarize` endpoint we talked about
* I updated the Context Manager downloads to call `get_metadata`, like Randy suggested
* I removed the aliases from the `/context/labels` response (so it now just returns a list of labels)

I'll just wait for the build to finish and then merge this in.

frontend/packages/portal-frontend/src/data-explorer-2/components/ContextManager/DownloadContextModal.tsx
```diff
@@ -47,18 +42,23 @@ function DownloadContextModal({
...
 
     fetchContext(contextHash)
-      .then(evaluateContext)
-      .then((evaluated) => {
-        let labels = evaluated.labels;
-
+      .then(fetchContextLabels)
+      .then(async (depmap_ids) => {
+        // For now, we assume that the downloads are always depmap_models
+        // And the actual labels need to be loaded with a separate request
```
rcreasi, line 48: This is not really true. It works with other types like genes but those other types don't have a concept of display name.

PR #26, opened 2024-08-08
Get dose table units

jessica-cheng:
Get rid of hardcoded units in the dose table and graph

Original dose table and X axis label with hardcoded units:
Screenshot 2024-08-08 at 10 20 20 AM

New dose table and X axis label with units retrieved from compound metadata
Screenshot 2024-08-08 at 10 22 14 AM


portal-backend/depmap/compound/models.py
```diff
@@ -239,7 +240,6 @@ class CompoundDose(Entity):
     )
 
     dose = Column(Float, nullable=False)
-    unit = "μM"
```
jessica-cheng, line 242: This doesn't seem like it was used for anything before so removing it to avoid confusion

frontend/packages/portal-frontend/src/compound/components/DoseResponseCurve.tsx
```diff
@@ -79,6 +80,7 @@ export class DoseResponseCurve extends React.PureComponent<DoseResponseCurveProp
     measurements: defaultPoints,
     curves: defaultParams,
     yUnits: "DEFAULT NOT FOUND",
+    xUnits: undefined,
```
pgm, line 83: I don't know what this is used for, but it looks like setting it to "DEFAULT NOT FOUND" would be consistent with what's being done for yUnits.

frontend/packages/portal-frontend/src/compound/components/DoseResponseTab.tsx
```diff
@@ -256,18 +256,26 @@ class DoseResponseTab extends React.Component<Props, State> {
...
+    let doseUnitsXLabel: string | undefined = undefined;
     const columns: Array<LongTableColumn> = this.state.doseResponseTable
       ? Object.keys(this.state.doseResponseTable[0])
           .filter(
             (col) => !["depmapId", "cell_line_display_name"].includes(col)
           )
-          .map((dose) => {
+          .map((doseWithUnits) => {
+            const splitDoseUnits = doseWithUnits.split(/\s/, 2); // units should be separated from dose by one whitespace
+            console.log(splitDoseUnits, doseWithUnits);
```
pgm, line 267: Good to drop this debugging log statement before merging in.

pgm (approved):

frontend/packages/portal-frontend/src/compound/components/DoseResponseTab.tsx
```diff
@@ -256,18 +256,26 @@ class DoseResponseTab extends React.Component<Props, State> {
...
+            const dose = splitDoseUnits[0];
+            const units = splitDoseUnits[1];
+            if (doseUnitsXLabel === undefined) {
```
pgm, line 270: Wrote about two concerns about this in a comment which appears to have not gotten saved... so I'll quickly try to rewrite that.

There are two things I'd typically flag as not-ideal: 
1. string parsing to get the units out of the label
2. `map()` is a functional programming idiom, but you have it creating a side effect (mutating `doseUnitsXLabel`) 

I don't believe either of these is a real problem but they are practices which I tend to discourage because they can lead to unexpected consequences when people make changes in the future. I'd say these are a little hacky but ok.

Anyway, there's even more surprising hacks in here already before you touched anything. For example:
```
 ${parseFloat(
              dose.replace("-", ".")	              dose.replace("-", ".")
            ).toPrecision(2)} μM`;	            ).toPrecision(2)} ${units}`            
```

I don't think my concerns are major, just encouraging avoiding those patterns in the future. I think a cleaner solution would be to change `portal-backend/depmap/compound/views/index.py` to include `units` as a separate field rather than append it onto the `dose` field.

jessica-cheng: That's a very good point! I think that passing `units` as a prop rather than a hacky string parsing solution is better for sure. I think I was confused initially because it seemed like this component derived its compound information from datasets rather than having the compound explicitly passed in as a prop but there's no reason why I can't change that!

jessica-cheng: The specific hack you mentioned that existed previously is indeed surprising but there's a note that explains somewhere why that's done on[ this file](https://github.com/broadinstitute/depmap-portal/blob/47284c500ee5389c8de75c952a6ddde6dda15150/portal-backend/depmap/compound/views/index.py#L391)

pgm (approved):
Looks great! Thanks!

PR #27, opened 2024-08-08
Remove workaround for Methylation public user uploads

snwessel:
[Asana](https://app.asana.com/0/1165651979405609/1207196143661714/f)

All of these special-case datasets have been moved to breadbox, so the workarounds aren't needed anymore 🎉 

I'm just going to make sure the tests pass and then merge this in.

### DIFF
--- a/portal-backend/depmap/interactive/config/categories.py
+++ b/portal-backend/depmap/interactive/config/categories.py
@@ -448,19 +448,3 @@ class PrismPoolConfig(CategoryConfig):
             return self.get_na_category().value
 
         return prism_pool_value_to_label[value]
-
-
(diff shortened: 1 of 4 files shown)


rcreasi:
Looks like this can go away now too:
https://github.com/broadinstitute/depmap-portal/blob/a7c2b421c0dce8868c4edbc42bf9d8854bb0432c/portal-backend/depmap/interactive/interactive_utils/get_and_process_data.py#L164-L165

snwessel:
Good point, thank you!

pgm (approved):
yay: 

3 removals of the word "Weird" and one removal of "Ugly" and one of "Special case"

PR #32, opened 2024-08-13
feat(breadbox): Added support for upload matrices in parquet format

pgm:
Adding a decent number of large datasets to breadbox takes a non-trivial amount of time. While waiting for data to load, I took the time to make changes to support uploading matrix datasets as parquet instead of CSV. (Parquet can be read/written much faster and is more compact, which should allow uploads to be somewhat faster)

Changes should be backwards compatible, and the api is only impacted by the addition of a new field `data_file_format` which can either be "csv" or "parquet" and will default to "csv".


breadbox/breadbox/schemas/dataset.py
```diff
@@ -119,6 +119,13 @@ class MatrixDatasetParams(SharedDatasetParams):
         ),
     ] = None
 
+    data_file_format: Annotated[
+        str,
```
jessica-cheng, line 123: Sounds like the value will be either "csv" or "parquet" only. How about we don't type it as `str` which could be any string and instead `Union[Literal["csv"], Literal["parquet"]]`?

jessica-cheng (approved):
LGTM!

breadbox/breadbox/schemas/dataset.py
pgm: Good point. I'll fix.

pgm:
messed up branch. recreating

PR #37, opened 2024-08-20
Cache resources

jessica-cheng:
[Asana task](https://app.asana.com/0/1190405531170073/1208047131211679)

### Cache Resources Page Responses
In order to show a post content, we must fetch each post's content from Discourse. However, we soon hit a rate limiting error from the Discourse API once we need to fetch anywhere between 10-14 posts which causes the entire page to crash. We aim to move away from an on-demand approach of getting the content to an on-deploy approach. This means, instead of the previous approach where we were fetching data from discourse everytime we hit the resources page view endpoint, we fetch the data every time we deploy. However, there are some issues with the on-deploy approach. The ops team wants to have full control over managing the Discourse posts' content. Moving to a truly on-deploy approach means that they are limited to seeing updated content every time the dev team deploys the app. The following changes were made to address the issues above:

- On the Client, we specify a mode "reload" to distinguish whether we fetch directly from Discourse and then store in the cache or straight from the cache
- The cache is implemented using SqliteDict (The original idea was to utilize sqlite-shelve. But sqlite-shelve library seems to have a read error where pickle is expecting the result to be bytes but the library is converting it into a string first. No raised issues about this could be found )
- A new endpoint `resources/reload` is created so that the ops team can reload the resources changes whenever a change is made
- A new command is added so that it can be called on deploying the app


jessica-cheng (commented):
@pgm With this solution in place, should I remove the flask view caching that we currently have for the resources page?

pgm:
> @pgm With this solution in place, should I remove the flask view caching that we currently have for the resources page?

Yes, I don't think the caching is needed if you're prefetching all the content.

portal-backend/depmap/cli_commands/db_load_commands.py
```diff
@@ -1761,3 +1763,28 @@ def _sync_metadata_to_breadbox():
...
+
+@click.command("reload_resources")
+@with_appcontext
+def reload_resources():
+    forum_api_key_value = current_app.config["FORUM_API_KEY"]
+    forum_url = current_app.config["FORUM_URL"]
+    if forum_api_key_value is None or forum_url is None:
+        raise Exception("Missing forum_api_key_value or forum_url values!")
+
+    if os.path.isfile(
```
pgm, line 1776: Refactor into a util method?

portal-backend/depmap/cli_commands/db_load_commands.py
```diff
@@ -1761,3 +1763,28 @@ def _sync_metadata_to_breadbox():
...
+            discourse_api_key = fp.read()
+    else:
+        discourse_api_key = forum_api_key_value
+
+    client = DiscourseClient(discourse_api_key, forum_url, True)
+    try:
+        refresh_all_category_topics(
+            client, current_app.config["FORUM_RESOURCES_CATEGORY"]
+        )
+    except Exception as e:
```
pgm, line 1789: try / except can be removed?

portal-backend/depmap/discourse/client.py
```diff
@@ -1,59 +1,130 @@
...
+from flask import current_app
+import os
+import json
+from sqlitedict import SqliteDict
 
 
 class DiscourseClient:
-    def __init__(self, api_key: str, base_url: str):
+    """The client has two modes:
+    1. In refresh mode, data is fetched from Discourse API and stored in a DB cache
```
pgm, line 13: Add a note that the "reload" parameter is controlling which of these two is in play?

portal-backend/depmap/discourse/client.py
```diff
@@ -1,59 +1,130 @@
...
+        url = "/categories.json"
+        with SqliteDict(self.resources_results) as db:
+            if self.reload:
+                res = self.get(url)
+                categories = res["category_list"]["categories"]
+                category = next(
+                    (c for c in categories if c["slug"] == category_slug), None
+                )
+                # Store response results
+                db[url] = category
```
pgm, line 76: key should be also a function of slug?

pgm (approved):
LGTM

portal-backend/depmap/public/views.py
```diff
@@ -9,6 +9,8 @@
 from depmap.public.resources import (
     get_root_category_subcategory_topics,
     create_sanitizer,
+    refresh_all_category_topics,
+    _read_forum_api_key,
```
rcreasi, line 13: A bit surprising to see an import with a leading underscore. Usually that means it's supposed to be private to the module.

jessica-cheng: We have multiple places in the portal where we import functions with a leading underscore but I agree that's not really the best practice. I had originally intended to use this function as part of public function but in the end decided not to. I'll rename this to not imply a private function though!

PR #39, opened 2024-08-22
Add endpoint to access private GCS files

jessica-cheng:
Granted Ops team ability to generate a signed url to give to users for large files. 

[See Asana task for remaining tasks to be done](https://app.asana.com/0/0/1208039245240536/1208102294494370/f)


portal-backend/depmap/private/views.py
```diff
@@ -0,0 +1,34 @@
...
+@blueprint.route("/<bucket_name>/<filename>")
+def portal_file_private_url(bucket_name, filename):
+    private_file_buckets = current_app.config.get("PRIVATE_FILE_BUCKETS")
+    # Make sure bucket provided is accepted in valid list of buckets
+    if private_file_buckets is None or bucket_name not in private_file_buckets:
+        abort(404)
+
+    try:
+        credentials = ServiceAccountCredentials.from_json_keyfile_name(
+            current_app.config["DOWNLOADS_KEY"]
```
jessica-cheng, line 28: I'm temporarily reusing existing credentials here. They seem to be specific to a project that I'm not planning on using, however, I do not want to use these credentials. I'm a bit confused however, on the process of how to create new credentials tho and would like help on that.

portal-backend/depmap/private/views.py
```diff
@@ -0,0 +1,34 @@
...
+        )
+        signed_url = sign_url(credentials, bucket_name, filename)
+    except Exception as e:
+        raise e
+    # Redirect to generated signed GCS url
+    return redirect(signed_url)
```
jessica-cheng, line 34: The current behavior for this is that on clicking this portal link, it automatically starts the download. I would like to open a new tab to view the file/video but there doesn't seem to be a flask function to do so. Any suggestions?

jessica-cheng: Also, this is just my own opinion of how I'd like this endpoint to work

portal-backend/depmap/private/views.py
pgm: I think we talked about this, and I think reusing credentials is fine.

Is this comment left over from before we talked?

portal-backend/depmap/private/views.py
```diff
@@ -0,0 +1,34 @@
...
+from oauth2client.service_account import ServiceAccountCredentials
+
+log = logging.getLogger(__name__)
+
+blueprint = Blueprint(
+    "private", __name__, url_prefix="/private", static_folder="../static"
+)
+
+
+@blueprint.route("/<bucket_name>/<filename>")
```
pgm, line 19: To support directories within the bucket (more specifically, to allow "/" in filename), I think this should be something like 

```
@blueprint.route("/<bucket_name>/<path:filename>")
```

portal-backend/depmap/private/views.py
```diff
@@ -0,0 +1,34 @@
...
+    # Make sure bucket provided is accepted in valid list of buckets
+    if private_file_buckets is None or bucket_name not in private_file_buckets:
+        abort(404)
+
+    try:
+        credentials = ServiceAccountCredentials.from_json_keyfile_name(
+            current_app.config["DOWNLOADS_KEY"]
+        )
+        signed_url = sign_url(credentials, bucket_name, filename)
+    except Exception as e:
```
pgm, line 31: Why catching an exception and rethrowing? I think this can be eliminated.

portal-backend/depmap/private/views.py
pgm: This might be easier to talk about. There are some options but let's discuss so I can make sure I'm on the same page.

pgm (approved):

portal-backend/depmap/private/views.py
jessica-cheng: Yes

portal-backend/depmap/private/views.py
jessica-cheng: Good point! The idea of creating directories within the bucket didn't cross my mind.

PR #40, opened 2024-08-22
update terminology used in Data Explorer 2

rcreasi:
I took (almost) every instance of the word "entity" and replaced it with "slice." It makes more sense to me but _is it worth it_? 🤔 

Are there other terms that should be given this treatment?


snwessel:
Thanks for working on this Randy! I don't know if I have particularly strong feelings one way or another about the change from `entity_labels` to `slice_labels` because it's not really moving things closer to the breadbox terminology. I am totally on board with it though if you think it's more helpful for describing things - and I think you made a good case for that at our last breadbox meeting.

My one ask would be that if we change the `entity_type` term, we change it to something like `dimension_type` or `slice_dimension_type` instead of just `slice_type`. I believe this `entity_type` value is expected to be one of the breadbox dimension types (ex. `depmap_model`, `gene`, `gene_pair`), right? If so, it would definitely be more helpful for my understanding to rename it to something that maps on to the breadbox terminology

frontend/packages/portal-frontend/src/data-explorer-2/query-string-parser.ts
```diff
@@ -46,18 +46,18 @@ const makeFeatureParser = (dimensionKey: DimensionKey) => (
     (d) => d.dataset_id === dataset_id
   );
 
-  const entity_type = dataset ? dataset.entity_type : "custom";
+  const slice_type = dataset ? dataset.slice_type : "custom";
```
snwessel, line 49: Oh I see now - I had been confused about what the difference was between a `slice_type` and a `dimension_type`, but it definitely makes sense to distinguish them if `slice_types` can also be "custom". Is this "custom" value also what you use when there is no dimension type (like for transient datasets / custom analysis results)?

snwessel (approved):
This looks great, thank you! This will definitely make it easier to migrate things to breadbox

frontend/packages/portal-frontend/src/data-explorer-2/query-string-parser.ts
rcreasi: `"custom"` is used like you're saying but we could've just as well used `null` for that. That's not really what `slice_type` is about.

`slice_type` was born when I realized I was using `entity_type` in two completely different ways:
- When it doesn't matter if it's a sample or feature (that's what `dimension_type` is for!)
  - (In the last round of changes I tried to fix all those) 
- When I wanted to keep track of what was being, well, sliced!
  - In other words, focusing on one row or column of a matrix so it can be plotted 

Consider this matrix:

|            |SOX10|TP53 |FOXA1|
|------------|-----|-----|-----|
| ACH-000001 | 0.1 | 0.1 | 0.1 |
| ACH-000002 | 0.2 | 0.2 | 0.2 |
| ACH-000003 | 0.3 | 0.3 | 0.3 |

`sample_type: "depmap_model"`
`feature_type: "gene"`

I could either slice it like this:
|            |SOX10|
|------------|-----|
| ACH-000001 | 0.1 |
| ACH-000002 | 0.2 |
| ACH-000003 | 0.3 |

or like this:
|            |SOX10|TP53 |FOXA1|
|------------|-----|-----|-----|
| ACH-000001 | 0.1 | 0.1 | 0.1 |

In the first case, my `slice_type` is `"gene"` and the columns form my possible `slice_labels`:
|SOX10|TP53 |FOXA1|
|-----|-----|-----|

In the second case, my `slice_type` is `"depmap_model"` and the rows form my possible `slice_labels`:
|            |
|------------|
| ACH-000001 |
| ACH-000002 |
| ACH-000003 |

It's basically an alternative to keeping track of `is_transpose` all over the place. It can be inferred from the `index_type` and `dataset_id` but it makes the code (at least the UI code) clearer when there's a name for it.

It should always be the case that `sample_type`/`feature_type` match exactly with `index_type`/`slice_type` or otherwise the two are swapped around.

rcreasi: Another way to think about this is to consider what you get back from the `/datasets_by_index_type` endpoint:
```js
{
  "depmap_model": [
    {
      "label": "CRISPR (Project Score, Chronos)",
      "dataset_id": "Chronos_Score",
      "slice_type": "gene",
      "index_type": "depmap_model"
    },
    // ...
  ],

  "gene": [
    {
      "label": "CRISPR (Project Score, Chronos)",
      "dataset_id": "Chronos_Score",
      "slice_type": "depmap_model",
      "index_type": "gene"
    },
    // ...
  ]
}
```
The same dataset is represented twice! This is because you can plot it in two different ways. In the UI, if you set your points to "models" it shows you the first list. If you set it to the "genes" it shows you the second list. Then when you pick something you want to plot, it necessarily has to be along the opposite axis. Again, that could be inferred but it's more convenient to just look it up.

snwessel: Ah, I see now! Thank you for the explanation!

portal-backend/depmap/data_explorer_2/views.py
```diff
@@ -462,18 +463,25 @@ def get_labels_matching_context():
     return make_gzipped_json_response(labels_matching_context)
 
 
+# TODO: Remove this endpoint. It's only used for one specific feature type
+# which is compound_experiment. The DE2 UI has a special case for that type
+# where it allows you to start by searching on a compound name alone. Then the
+# dataset name is used to resolve ambiguities in the relatively small number of
+# cases where there's more than one experiment per compound. In the future, we
+# can solve this a manner more consistent with other types by leveraging the
+# search indexer.
```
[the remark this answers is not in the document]
snwessel: Also, since I'm trying to keep track of everything that needs to be done for Elara V2, I added a task for this to Asana: https://app.asana.com/0/1165651979405609/1208160165634335/f

PR #43, opened 2024-08-23
refactor(breadbox): Move DE2 Context and Slice ID utils to shared module

snwessel:
[Asana](https://app.asana.com/0/1165651979405609/1207979348562841/f) 

In order for Breadbox to support DE2's context endpoints, it needs to have access to a few utils:
* `ContextEvaluator` takes a context definition. This class has a function that tells you whether a given feature label is a match for that context. It also does some caching. 
* `decode_slice_id` takes a slice id string and breaks it into its respective parts. 

I am moving those utils to a shared module (`depmap_compute`) so they can be used by breadbox. I also added some documentation and typing.

I also defined a couple of schema classes that will used in the context endpoints I'm working on adding to breadbox. 

**My next steps will be to:**
* Update breadbox to support sample lookup by label
* Define the context endpoints in breadbox


depmap-compute/depmap_compute/context.py
```diff
@@ -101,3 +112,30 @@ def walk(node, key):
...
+    """
+    Originally based on the function of the same name from vector_catalog.SliceSerializer,
+    Data Explorer 2 slice ids are a superset of the legacy slice ids.
+    Originally, slice IDs were formatted like "slice/some_dataset_id/some_feature_label/label", 
+    or "slice/some_dataset_id/some_feature_id/entity_id", where the last part of the string
+    (originally called the SliceRowType) specifies whether the feature is being identified by ID or by label. 
+
+    The DE2 slice IDs give you flexibility by letting you query samples as well using the "transpose_label" specifier.
+    When "transpose_label" is used as the last segment of the slice ID, it means we should query for samples.
+    """
```
snwessel, line 127: I tried to add some better documentation to this `decode_slice_id` function because I am often confused about which slice IDs formats are supported in which parts of the codebase. I do think there is still some work remaining to clean this up, and I think there needs to be a bit more discussion about what we want from slice IDs - especially since they're stored in the DE1 and DE2 URLs, so they're difficult to change once they're in use. 

As an example of why I find this confusing, DE1 supports the slice types "label" and "entity_id", while DE2 supports "label" and "transpose_label". It seems like someday, we may want DE2 on breadbox to be able to look up features or samples by ID as well, which would mean we'd need four "slice types":
* label
* entity_id
* transpose_label
* transpose_id (?)

I'll add a note to our Breadbox discussion list to talk about this more.

snwessel:
I see that the pyright-ratchet step is failing because it can't import my new `depmap-compute` file, but I don't really have a way of fixing that without publishing a new version of the package - and I don't want to randomly publish a new version because that might interfere with the automated version bumping that happens when the PR is merged to master 😪 

So I am just going to let the checks fail here - I have made sure that all the tests are passing locally.

breadbox/breadbox/schemas/context.py
```diff
@@ -0,0 +1,15 @@
...
+    num_candidates: int
+    num_matches: int
+
+
+class Context(BaseModel):
+    # Context expression examples:
+    # - { "!": [ { "var": "slice/lineage/1/label" }, "Breast" ] }
+    # - { "==": [ { "var": "entity_label"}, "ACH-000001" ] }
+    # - { "==": [ {"var": "slice/growth_pattern/all/label"}, "Adherent" ] }
+    expr: dict[str, list]
```
rcreasi, line 14: `expr` can also be a bool. That makes this special "All" option trivial to encode (`expr == True`).

all

[https://jsonlogic.com/#always-and-never](https://jsonlogic.com/#always-and-never)

portal-backend/tests/depmap/data_explorer_2/context/test_custom_operators.py
```diff
@@ -75,6 +77,6 @@ def expressions_are_equivalent(boolean_value, json_logic_expr):
 
     # These expressions don't use variables (just pure logic)
     var_name = "dummy variable"
-    result = ContextEvaluator(context).is_match(var_name)
+    result = ContextEvaluator(context, slice_to_dict).is_match(var_name)
```
rcreasi, line 80: It really doesn't matter but `lambda: None` would work in place of `slice_to_dict` I think.

This test is a little funny. It wants to test the internal logic engine but that's not exposed so I had to pass a meaningless variable that never gets looked up.

depmap-compute/depmap_compute/context.py
```diff
@@ -101,3 +112,30 @@ def walk(node, key):
...
+        parts[0] == "slice" and len(parts) >= 4 and len(parts) <= 5
+    ), f"Malformed slice_id: {slice_id}"
+
+    if len(parts) == 5:
+        # handle dataset IDs with slashes in them
+        parts[1:3] = ["/".join(parts[1:3])]
+
+    dataset_id = unquote(parts[1])
+    dimension_identifier = unquote(parts[2])
+    slice_type = unquote(parts[3])  # "label", "entity_id", or "transpose_label"
```
rcreasi, line 139: The original implementation used an enum for this:
https://github.com/broadinstitute/depmap-portal/blob/53bee77ccceedde16cd2efcde57581034bbca3ac/portal-backend/depmap/vector_catalog/models.py#L13-L15

I didn't do the same because I didn't want the hack to seem "official." But like we were saying above, DE2 has been out in the wild long enough that it's de facto official now.

depmap-compute/depmap_compute/context.py
```diff
@@ -59,10 +63,17 @@ def is_match(self, entity_label):
 # But we don't need to "perfectly" override it; just well enough to trick the
 # JsonLogic library.
 class _LazyContextDict(dict):
-    def __init__(self, context_type, entity_label, cache):
+    def __init__(
+        self,
+        context_type: str,
+        entity_label: str,
+        cache: dict,
+        get_slice_data: Callable[[str], dict[str, Any]],
```
rcreasi, line 71: This looked funny at first but now I think I like it. Dependency injection FTW!

breadbox/breadbox/schemas/context.py
snwessel: Oh, thank you! That's good to know. I'll update the typing now

depmap-compute/depmap_compute/context.py
snwessel: Thanks, I agree it would be nice to have this defined by an enum now that it's more official. 

I wasn't sure if there are other slice identifier types we want to support going forward (for example, do we want to be able to look up both features and samples by ID? or neither?) - so I added a note to the breadbox discussion list. If we all agree about that, I'll define this as an enum and add some better docs 👍 
https://app.asana.com/0/1206925945720127/1208125648857065/f

depmap-compute/depmap_compute/context.py
```diff
@@ -36,15 +37,18 @@ def __init__(self, context):
         self.context_type = context["context_type"]
         self.expr = _encode_dots_in_vars(context["expr"])
         self.cache = {}
+        self.get_slice_data = get_slice_data
 
-    def is_match(self, entity_label):
+    def is_match(self, entity_label: str):
```
rcreasi, line 42: Once I merge my changes you'll have some conflicts. I renamed `entity_label` -> `dimension_label` to keep things a little more consistent.

PR #45, opened 2024-08-23
fix(breadbox): Fix pydantic errors preventing generation of get dataset data methods in client

jessica-cheng:
Fix pydantic errors preventing generation of get dataset data methods in client


snwessel (approved):
LGTM!

PR #46, opened 2024-08-26
Fix appearance of retracted files on downloads page

alimourey:
Fix retracted files so that they are labeled with "(retracted)", do not have a downloads button, and display the retraction_override text about the description. Related Asana task: https://app.asana.com/0/inbox/1200970789241572

Example of a file set with retracted files on Skyros.

Screenshot 2024-08-26 at 6 22 55 PM


pgm (approved):
Looks great

PR #52, opened 2024-09-03
Add consistent label support to the data_access interface

snwessel:
Currently, the way the `portal-backend` uses sample labels doesn't match the way we configure and use sample labels in breadbox. 

For example:
* In breadbox, `depmap_models` are configured [here](https://github.com/broadinstitute/depmap-deploy/blob/08651f06cb2994318ba6a5293bcdcab046483485/depmap_deploy/breadbox_loader/configs/release_files.py#L20C60-L20C72) to use `CellLineName` as their labels. 
* However, endpoints like DE2's `/context/labels` ignore this config and continue to return depmap_ids as the "labels" instead of using cell line names. 

Eventually, we would like DE2 to be able to use whatever "sample label" is configured in breadbox. This PR adds/updates some methods in the `data_access` interface to make that possible. 

**This includes:**
* Defining`get_dataset_sample_labels_by_id`
* Replacing the little-used function `get_dataset_feature_ids_by_label` with `get_dataset_feature_labels_by_id` (for consistent naming)

Once we decide on a new slice ID format, I will implement any other data_access functions that are needed to support the new format (ex. looking up data by sample label).

### DIFF
--- a/portal-backend/depmap/data_access/__init__.py
+++ b/portal-backend/depmap/data_access/__init__.py
@@ -3,7 +3,7 @@ from .interface import (
     get_all_matrix_dataset_ids,
     get_all_matrix_datasets,
     get_matrix_dataset,
-    get_dataset_feature_ids_by_label,
+    get_dataset_feature_labels_by_id,
     get_dataset_feature_labels,
     get_dataset_sample_ids,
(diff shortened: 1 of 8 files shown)


pgm (approved):
LGTM

PR #53, opened 2024-09-03
refactor(breadbox): Remove catalog nodes

snwessel:
#### This is the last step in removing catalog nodes from breadbox 🎉 

[Asana](https://app.asana.com/0/1165651979405609/1206585568236568/f) 

This ended up being pretty simple since we've already removed all catalog-node-dependent functionality. My process for this last step was to:
* Replace all of the `CatalogNode` references in tests (usually by querying `TabularColumn` instead), and then confirm the tests still pass
* Remove `CatalogNode` references from the rest of breadbox. At this point, it was mostly just places where CatalogNodes were being inserted or deleted from the database. 
* Drop the `CatalogNode` table from the database.


rcreasi:
Is this comment still true?
https://github.com/broadinstitute/depmap-portal/blob/f2a8a79aba451297a4f9c50184f3f395c6b070b6/breadbox/breadbox/models/dataset.py#L201-L203

snwessel:
@rcreasi good question! It looks like that is being used more regularly now, so I just updated the comment:
```
# Denormalized data: this information is also stored in the dataset's corresponding 
# feature/sample/index type name column (the information is duplicated here for convenience)
dataset_dimension_type = Column(String, nullable=True)  
```

pgm (approved):

jessica-cheng (approved):
LGTM!

PR #57, opened 2024-09-09
fix(breadbox): throw FileValidationError for duplicate CSV columns

snwessel:
[Asana](https://app.asana.com/0/1165651979405609/1208078862632786/f)

When the pandas `read_csv` function encounters duplicate columns, it renames one of the duplicates instead of throwing an error. As a result, our CSV reading function needs to do one additional step:
1. First, read the header with `csv.DictReader` and check for duplicate columns
2. Then, read the full file with pandas and continue on as usual. 

I learned that this isn't an issue for parquet file uploads because the file format doesn't support duplicate column names. If you try to do `pd.to_parquet` with a duplicate column name, you get a clear error message like: `ValueError: Duplicate column names found: ['a', 'b', 'b']`


jessica-cheng (approved):
LGTM!

PR #58, opened 2024-09-09
add flag for faster development

rcreasi:
This adds a new `yarn dev:portal:nocheck` script that will start the frontend dev server with type checking disabled. This is a nice option to use when you're making lots of small incremental changes and you prefer speed over visibility of errors. You'll notice updates happening faster but you'll also no longer see type errors in the terminal or browser console.

Modern IDEs like Visual Studio Code run their own language server which is constantly compiling the codebase in the background in order to show inline error messages. This makes those other forms of error reporting somewhat redundant.

Note that this only affects the TypeScript compiler. ESLint will continue to run as before so you'll still see linting errors.

### DIFF
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -43,8 +43,17 @@ or
 yarn dev:portal
 ```
 
-(But you can still do `./flask webpack` from the portal-backend directory if
-that's what you're used to. All it does is call `yarn dev:portal` for you).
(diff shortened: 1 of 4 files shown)


jessica-cheng (approved):
This is so much faster! Thank you so much! 🎉

PR #59, opened 2024-09-09
feat(breadbox): Add display name field to Dimension Type

jessica-cheng:
Add `display_name` field to dimension type table and as a parameter to POST endpoint
Some guesswork for display name done to fill out a value for the dimension type and not break things


breadbox/breadbox/models/dataset.py
```diff
@@ -39,6 +39,14 @@ def __init__(self, enum: Type[T], **kwargs: Any) -> None:
 
 import enum
 
+# context-sensitive default function
+def default_display_name(context):
```
pgm, line 43: Is this here to auto-populate display_name in the event it's None in an effort to offer backwards compatibility?

My experiences have lead me to be very cautious about coercing data at a low level (in this case, using sqlalchemy to fill in values). More specifically, when I say, "my experiences" I mean I've had to debug weird issues where automatically populated fields interact in surprising ways and/or it's less obvious that some logic is running to get a value.

This case is simple enough, I think it's realistically fine, but I have a bias towards putting coercions like this as early in the call chain as possible so that everything is consistent from beyond the point of validating and parsing inputs. 

As it stands, I see that `AddDimensionType` has display_name as type `str` and not `Optional[str]`. 

Is there support for display_name being omitted when someone adds a type? (to keep it backwards compatible)

jessica-cheng: Yes, this is auto-populate `display_name` function is an effort in backwards compatibility. This has mainly been helpful when making sure the tests were backwards compatible initially. 

I agree that this now seems confusing since I've made `display_name` non-optional and as we've discussed earlier, there are other repos that are using the function to add a dimension type so backwards compatibility would be necessary here. I've made the change now to make this request parameter optional

PR #61, opened 2024-09-10
fix(breadbox): added job to check breadbox client

pgm:
I've hit twice now cases where method broke in breadbox_client due to the auto-generated code changing. It's a bit of a pain to set up automated tests for breadbox_client because it requires running an instance of breadbox. Maybe worthwhile eventually, but these two errors were things that pyright could catch.

Trying to at least put in a pyright check to confirm things look sane.

(And also put in fixes to satisfy pyright)


breadbox-client/breadbox_facade/client.py
```diff
@@ -368,134 +368,12 @@ def get_dimension_type(self, name: str):
...
-            metadata_file=metadata_file,
-            taiga_id=taiga_id,
-            annotation_type_mapping=type_mapping,
-            id_mapping=IdMapping.from_dict(id_mapping) if id_mapping else UNSET,
-            properties_to_index=properties_to_index if properties_to_index else UNSET,
-        )
-        breadbox_response = add_feature_type_client.sync_detailed(client=self.client, multipart_data=params)
-        return self._parse_client_response(breadbox_response)
-
-    def add_sample_type(
```
snwessel, line 399: Heads up, these old functions are still being used by the old `sync_metadata_to_breadbox` utility, so these deletions may break the `portal-backend` type checker (once the portal starts referencing this new version of the breadbox client). 

It seems like that `sync_metadata_to_breadbox` process has been broken for a little while now (I was trying to use it this afternoon and was disappointed),  and I'm surprised no one else has complained about it, so I added a topic to our breadbox discussion list https://app.asana.com/0/1206925945720127/1208261191969013/f

snwessel (approved):
Woohoo! 🎉  This was much needed, thank you!

PR #62, opened 2024-09-11
Updated recreate_dev_db to not use any deprecated API calls

pgm:
Upgraded the breadbox client used by the portal and updated sync_metadata_to_breadbox to avoid using any deprecated endpoints.


snwessel (approved):
Thank you, this is so helpful! Looks good 👍

PR #64, opened 2024-09-11
feat(breadbox): Update dimension type display name and allow partial updates

jessica-cheng:
Add parameter `display_name` to PATCH endpoint for updating dimension type and allow for partial updates meaning only fields that you want to be updated need to be included in the request parameters


pgm (approved):
LGTM

PR #66, opened 2024-09-12
chore(breadbox): Deprecate feature type and sample type endpoints

jessica-cheng:
These endpoints should now be under the URI "dimensions"


snwessel (approved):
Looks good, thanks for updating this!

PR #68, opened 2024-09-12
Vector catalog fixes + interactive module refactoring

snwessel:
**The issue I'm trying to address:**
* Some datasets are defined in both breadbox and the legacy portal (ex. `Chronos_Combined`)
* For some portal features like DE2, we want to return the breadbox version when we receive a query for a dataset ID like `Chronos_Combined` (assuming this ID is configured as a breadbox "dataset given id")
* Other portal features like VectorCatalog will break if we start returning values for the breadbox version of the dataset, so we need to keep returning the old version. 

**The solution:** 
The `interactive` and `compute` modules don't really need to use the `data_access` interface because they call the `breadbox_shim` to load breadbox data. Going forward:
1. Parts of the portal which do not support breadbox datasets should be clearly labeled.
2. Those parts of the portal should use the old `interactive_utils` interface instead of the new `data_access` interface.
3. There will still be a couple of places in these modules (`interactive` and `compute`) were it _is_ actually helpful to use the data_access interface, but those places should be clearly labeled as having support for breadbox datasets. 

**Next steps:** 
1. Look through the cell line, gene, and compound pages to check whether they need to be updated at all to work with the new given ID functionality. Edit: just checked and they seem fine
2. Follow up on my other PR to change the default behavior to querying breadbox for given IDs: https://github.com/broadinstitute/depmap-portal/pull/67

### DIFF
--- a/portal-backend/depmap/breadbox_shim/breadbox_shim.py
+++ b/portal-backend/depmap/breadbox_shim/breadbox_shim.py
@@ -5,6 +5,7 @@ from typing import Any, Optional, Union
 from breadbox_client.models.compute_response import ComputeResponse
 from breadbox_client.models import (
     MatrixDatasetResponse,
+    MatrixDatasetResponseFormat,
     TabularDatasetResponse,
     FeatureResponse,
     ValueType,
@@ -34,7 +35,13 @@ class BreadboxVectorCatalogChildNode(Node):
     for breadbox nodes, so a simpler version is being used as a return type in the shim.
     """
 
-    def __init__(self, slice_id: str, label: str, is_terminal: bool):
+    def __init__(
+        self,
+        slice_id: str,
+        label: str,
+        value: str,  # For datasets, this should be the ID, for features, it's the label
+        is_terminal: bool,
+    ):
         # Fill in defaults for values that are required in the Node type
         super().__init__(
             id=slice_id,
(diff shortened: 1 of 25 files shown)


portal-backend/depmap/interactive/views.py
```diff
@@ -145,12 +147,6 @@ def get_datasets():
         key=lambda dataset: data_access.get_sort_key(dataset["value"]),
     )
 
-    # append breadbox datasets to the end of the list
-    for dataset_node in breadbox_shim.get_dataset_nodes(catalog_type="continuous"):
-        combined_datasets.append(
```
snwessel, line 150: This had been causing the breadbox datasets to be returned in duplicate by this endpoint

portal-backend/depmap/breadbox_shim/breadbox_shim.py
```diff
@@ -34,15 +35,21 @@ class BreadboxVectorCatalogChildNode(Node):
...
+    ):
         # Fill in defaults for values that are required in the Node type
         super().__init__(
             id=slice_id,
             key=None,
             attrs={},
             is_terminal=is_terminal,
             label=label,
-            value=label,
+            value=value,  # maps onto childValue in the API response
```
snwessel, line 52: This change fixes an issue with the "View Results in DE2" button, which reads from this `"childValue"` field.

pgm (approved):
At a high level, I think I understand what's going on here and looks good to me. I like that data_access has shrunk and now all vector catalog tree is back to making direct calls to interactive config. I think that's a good place to draw the line.

PR #69, opened 2024-09-12
Set default resources topic

jessica-cheng:
Ops has decided that they want to draft their own post that the resources page will default to when a user first arrives at the resources page. I've added a new Flask variable to store the stable topic identifier from Discourse that maps to the default post that ops wants to show.

### DIFF
--- a/config/public/settings.py
+++ b/config/public/settings.py
@@ -55,6 +55,7 @@ class ExternalConfig(RemoteConfig):
     RELEASE_NOTES_URL = "https://forum.depmap.org/c/announcements/15"
     FORUM_URL = "https://forum.depmap.org/"
     FORUM_RESOURCES_CATEGORY = "resources-prototype"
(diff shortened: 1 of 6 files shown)


config/public/settings.py
```diff
@@ -55,6 +55,7 @@ class ExternalConfig(RemoteConfig):
     RELEASE_NOTES_URL = "https://forum.depmap.org/c/announcements/15"
     FORUM_URL = "https://forum.depmap.org/"
     FORUM_RESOURCES_CATEGORY = "resources-prototype"
+    FORUM_RESOURCES_DEFAULT_TOPIC_ID = 3396
```
jessica-cheng, line 58: This maps to [this topic](https://forum.depmap.org/t/welcome/3396) . I don't believe it's an issue to show the identifier here even though this public forum topic is currently not available for public view yet.

pgm: Yeah, I don't see a problem

PR #71, opened 2024-09-16
Style resources page

jessica-cheng:
[Design](https://www.figma.com/design/YYEB9MnU8CYPdCyytvOhRi/Resources-Page?node-id=0-1&node-type=canvas&t=JNCjxzuuZeGVg9hN-0)

https://github.com/user-attachments/assets/c9911011-cda2-4c7a-b0a0-3c0f13cbbef7

### DIFF
--- a/portal-backend/depmap/static/css/public/resources.scss
+++ b/portal-backend/depmap/static/css/public/resources.scss
@@ -1,9 +1,11 @@
 .depmap-accordion {
   margin: 0px !important;
   .accordion-header {
+    padding: 8px 20px 8px 40px !important;
     background-color: #f5f8fd !important;
+    border-bottom: 1px solid #585858 !important;
     .accordion-title {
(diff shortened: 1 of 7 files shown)


portal-backend/depmap/static/css/public/resources.scss
```diff
@@ -0,0 +1,12 @@
...
+  .accordion-header {
+    padding: 8px 20px 8px 40px !important;
+    background-color: #f5f8fd !important;
+    border-bottom: 1px solid #585858 !important;
+    .accordion-title {
+      font-weight: 500;
+      font-size: 18px;
+    }
+  }
+}
```
jessica-cheng, line 12: These styles override the Accordion component's `depmap-accordion` style. I feel like this could be somewhat simplified if the Accordion component took in style as props but I was unsure how this would affect all the other places in the portal that used it.

PR #74, opened 2024-09-18
Resources open init post accordion

jessica-cheng:
We want the accordion open for the initial post on the resources page. However, some changes to the Accordion component were necessary since it appears that the toggle behavior is somewhat buggy

The previous buggy behavior was that if the initial Accordion is set to be open, when we choose a different post, the accordion closes.

https://github.com/user-attachments/assets/3521c5a7-3ad5-4cfd-88c0-1fd770916410

Upon further investigation, it seems like this is due to the Accordion component's `isOpen` state. When that is set to `isOpen=true`, the next time that Accordion's state is changed it will automatically be set to `isOpen=false` through its `toggle()` function. However, I only want to activate the toggle function when we click the Accordion and in order to do that, it needs to determine the previous state. Note that the default `isOpen` state is `undefined`

https://github.com/user-attachments/assets/4602c6d3-b4fa-4680-9c59-73eba06d158f

### DIFF
--- a/frontend/packages/@depmap/interactive/src/components/Accordion.tsx
+++ b/frontend/packages/@depmap/interactive/src/components/Accordion.tsx
@@ -42,7 +42,7 @@ export default class Accordion extends React.Component<Props, State> {
     const { isOpen } = this.props;
 
     if (prevProps.isOpen !== isOpen) {
-      if (isOpen !== undefined) {
+      if (isOpen !== undefined && prevProps.isOpen !== undefined) {
         this.toggle();
       }
     }
--- a/frontend/packages/@depmap/interactive/src/components/Accordion.tsx
+++ b/frontend/packages/@depmap/interactive/src/components/Accordion.tsx
@@ -42,7 +42,9 @@ export default class Accordion extends React.Component<Props, State> {
     const { isOpen } = this.props;
 
     if (prevProps.isOpen !== isOpen) {
-      this.toggle();
+      if (isOpen !== undefined) {
+        this.toggle();
+      }
     }
   };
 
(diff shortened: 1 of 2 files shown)


jessica-cheng (commented):
@alimourey @rcreasi It appears you both are most familiar with this component which is why I've set you both as reviewers. I tried to make sure other places that use the Accordion component weren't affected but it is possible I may have missed something

frontend/packages/@depmap/interactive/src/components/Accordion.tsx
```diff
@@ -42,7 +42,9 @@ export default class Accordion extends React.Component<Props, State> {
     const { isOpen } = this.props;
 
     if (prevProps.isOpen !== isOpen) {
-      this.toggle();
+      if (isOpen !== undefined && prevProps.isOpen !== undefined) {
+        this.toggle();
+      }
```
alimourey, line 47: If I'm understanding correctly, I think this is very similar to something I tried to do at one point. @rcreasi  suggested a different hack that might be safer and worked quite well for me on the Data Page: https://github.com/broadinstitute/deprecated-depmap-repo/pull/717#discussion_r1586423492

frontend/packages/portal-frontend/src/resources/components/ResourcesPage.tsx
```diff
@@ -67,7 +101,9 @@ export default function ResourcesPage(props: ResourcesPageProps) {
                       <ListGroupItem
```
alimourey, line 101: Another thing to check. Are these links triggering the Accordion's internal onClick event?

pgm:
This one is on hold due to trying a different approach.

PR #76, opened 2024-09-27
Support v2 Contexts in portal backend

snwessel:
[Asana](https://app.asana.com/0/1165651979405609/1208261191969003/f)

The main goal here is to support "version 2" contexts described in Randy's proposal [here](https://docs.google.com/document/d/1wQNP4YfcI27NL9imBbb_1ie7L6koCE8_wXOoW2P-Gk8/edit). There is still some work to be done on this, but I wanted to get something functional and ask for feedback earlier rather than later. 

**A couple of caveats:** 
* These new `v2` endpoints are currently _only_ able to accept the new context format. I am thinking it will probably be helpful/necessary to have this endpoint accept _either_ format, but I wanted to ask first if that's in line with what the frontend expects. 
* This is not 100% finished. It only currently supports 3/5 identifier types: `sample_id`, `feature_label`, and `column`. My next step will be to add support for `sample_label` and `feature_id` lookups. 

**Along the way, I also:**
* Tried to improve naming and documentation
* Wrote a bunch of tests to make sure `ContextEvaluator` is working as expected

### DIFF
--- a/depmap-compute/depmap_compute/context.py
+++ b/depmap-compute/depmap_compute/context.py
@@ -36,6 +36,7 @@ class ContextEvaluator:
         """
         self.context_type = context["context_type"]
         self.expr = _encode_dots_in_vars(context["expr"])
+        # Cache is keyed by Slice ID. Each value is an entire dictionary of slice values.
         self.cache = {}
         self.get_slice_data = get_slice_data
 
@@ -46,23 +47,29 @@ class ContextEvaluator:
         the expression, including any variables ("var" subexpressions) which
         are bound by using a magic dict that does lookups lazily.
         """
-        data = _LazyContextDict(
+        dictionary_override = _LazyLoadingSliceLookup(
             self.context_type, dimension_label, self.cache, self.get_slice_data
         )
 
         try:
-            return jsonLogic(self.expr, data)
+            return jsonLogic(self.expr, dictionary_override)
         except (TypeError, ValueError) as e:
             print("Exception evaluating", self.expr, "against", dimension_label)
             print(e)
             return False
 
 
-# Interesting thread on overriding the Dict class:
-# https://stackoverflow.com/questions/3387691/how-to-perfectly-override-a-dict
-# But we don't need to "perfectly" override it; just well enough to trick the
-# JsonLogic library.
-class _LazyContextDict(dict):
+class _LazyLoadingSliceLookup(dict):
+    """
+    The JsonLogic library wants to be passed a dictionary of values. However, we need to 
+    inject our own special cases and caching, so we override the dictionary class with 
+    special functionality. Interesting thread on overriding the Dict class:
+    https://stackoverflow.com/questions/3387691/how-to-perfectly-override-a-dict
+    But we don't need to "perfectly" override it; just well enough to trick the JsonLogic library.
+
+    This implementation uses and updates the cache that's been passed in to the constructor.
+    """
+
     def __init__(
         self,
         context_type: str,
@@ -76,6 +83,10 @@ class _LazyContextDict(dict):
         self.get_slice_data = get_slice_data
 
     def __getitem__(self, prop):
+        """
+        Given a slice ID, get the slice value which corresponds to the 
+        "dimension_label" that's already been passed into the constructor of this class.
+        """
         # Handle trivial case where we're just looking up a dimension's own
         # label. Note that this is called "entity_label" for historical
         # reasons.
@@ -87,6 +98,8 @@ class _LazyContextDict(dict):
                 self.cache[prop] = self.get_slice_data(prop)
 
             return self.cache[prop][self.dimension_label]
+
+        # TODO: handle new-style slice IDs
 
         raise LookupError(
             f"Unable to find context property '{prop}'. Are you sure a corresponding "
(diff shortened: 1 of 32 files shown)


portal-backend/depmap/data_explorer_2/views.py
```diff
@@ -464,6 +470,64 @@ def get_labels_matching_context():
...
+@csrf_protect.exempt
+def evaluate_v2_context():
+    """
+    Get the full list of labels (in any dataset) which match the given context.
+    """
+    inputs = request.get_json()
+    context = inputs["context"]
+    dimension_type = context.get("dimension_type")
+    if dimension_type is None:
+        abort(400, "v2 Contexts must have a 'dimension_type' field")
```
snwessel, line 483: @rcreasi these couple of lines differ from the proposal you put together so I wanted to call them out and get your opinion. The old context requests had a field called `"context_type"` which really stores what breadbox calls the "dimension type". My thinking was that:
1. this might be a good opportunity to rename a confusing field name
2. changing this gives us an easy way to check "is this in the old or new format?"

If you have opposition to it, let me know and I'd be happy to change it back 👍

portal-backend/depmap/data_explorer_2/views.py
```diff
@@ -464,6 +470,64 @@ def get_labels_matching_context():
...
+    context_evaluator = ContextEvaluator(context, data_access.get_slice_data)
+    ids_matching_context = []
+    labels_matching_context = []
+    for given_id, label in all_labels_by_id.items():
+        if context_evaluator.is_match(given_id):
+            ids_matching_context.append(given_id)
+            labels_matching_context.append(label)
+
+    return make_gzipped_json_response(
+        {"ids": ids_matching_context, "labels": labels_matching_context,}
```
snwessel, line 498: This now returns _both_ labels and IDs for anything that matches the context

portal-backend/depmap/data_explorer_2/views.py
rcreasi: I'm sure this will come back to bite me, just because the old name exists in a lot of places, but I think it makes sense rename it. They are truly synonyms -- a `context_type` isn't anything special. It's there to make sure it's compatible with the dimension you're working with so it should've always been called `dimension_type`.

rcreasi: The only reason I could think of to leave the name distinct is that the `dimension_type` might be `"screen"` and it could theoretically  make sense (though there is no support for this kind of thing today) to apply a `"depmap_model"` context against it.

rcreasi: But even then that would be two dimension types that have a special relationship. So there's really no reason to call it `context_type`.

portal-backend/depmap/data_explorer_2/views.py
rcreasi: That's cool. Elsewhere I noticed you were removing `make_gzipped_json_response()` and just returning the dict. The more of that we do, the more motivated I'll be to [move the compression to the nginx](https://app.asana.com/0/1200435587978125/1201078871707042/f) where it belongs :)

rcreasi: That gzip helper doesn't respect the request headers which really confused Phil when he was trying to use `curl` on these endpoints.

depmap-compute/depmap_compute/context.py
```diff
@@ -28,14 +31,128 @@
...
+        # Takes a slice query, returns a dictionary of slice values (indexed by ID)
+        self.get_slice_data = get_slice_data
+
+        # The cache is used so that slice values only need to be looked up once per context.
+        # - The keys in this dictionary are slice queries encoded as a tuple
+        #   (ex. `("Chronos_combined", "SOX10", "feature_label")`).
+        # - The values are each an entire dictionary of slice values (indexed by given ID)
+        self.cache = {}
+
+    def is_match(self, given_id: str):
```
rcreasi, line 60: It doesn't have to be part of this PR. I liked the idea you had of passing a list of IDs and getting back a filtered list. Something like:
```
  def evaluate(self, given_ids: List[str]): List[str]
```
Or at that point this really wouldn't have to be class at all and could just be an `evaluate_context()` function.

depmap-compute/depmap_compute/context.py
```diff
@@ -28,14 +31,128 @@
...
+    For context examples, see tests.
+    """
+
+    def __init__(
+        self, context: dict, get_slice_data: Callable[[SliceQuery], pd.Series],
+    ):
+        """
+        A `context` dict should have:
+            - a `dimension_type` such as "depmap_model" (which is not used in this evaluator)
+            - an `expr` such as { "==": [ { "var": "slice/lineage/1/label" }, "Breast" ] }
```
rcreasi, line 46: It's worth mentioning there's now a `"vars"` key as well.

depmap-compute/depmap_compute/context.py
```diff
@@ -28,14 +31,128 @@
...
+        get_slice_data: Callable[[SliceQuery], pd.Series],
+        slice_query_vars: dict[str, dict[str, str]],
+    ):
+        self.given_id = given_id
+        self.get_slice_data = get_slice_data
+        # The cache is stored outside of this class so it can be reused.
+        self.cache = cache
+        self.slice_query_vars = slice_query_vars
+
+    def __getitem__(self, var_name: Union[str, dict]) -> Any:
```
rcreasi, line 108: I like calling this `var_name` but it's always a string. Maybe you were confused because it acts a key for `slice_query_vars` which is a dict?

depmap-compute/depmap_compute/context.py
```diff
@@ -28,14 +31,128 @@
...
+        """
+        Given a variable from the context definition, load the corresponding slice value. 
+        Look up the value by the "given_id" that's already been passed into the constructor of this class.
+        Context vars can be either:
+        - a slice query (used to reference a dimension value)
+        - the string "given_id" (used to reference an id)
+        """
+        # There is a special case where "given_id" may be specified instead of
+        # a slice query. This allows our context definitions to reference ids, which
+        # wouldn't otherwise be possible because slice queries are used to load dataset values.
```
rcreasi, line 118: I appreciate the additional comments clarifying this weird magic 💯

rcreasi (approved):

depmap-compute/depmap_compute/context.py
snwessel: Oh good point! Yes, I was originally thinking this would take the slice query itself, and I forgot to update the type when I changed it to take the var_name instead.

PR #85, opened 2024-10-11
Dimension type form

jessica-cheng:
Screenshot 2024-10-10 at 2 23 54 PM

Screenshot 2024-10-15 at 12 06 11 PM

Also fixed issue where null selections were causing error:
Screenshot 2024-10-10 at 2 08 14 PM


frontend/packages/@depmap/api/src/ApiContext.tsx
```diff
@@ -105,6 +109,16 @@ export interface SharedApi {
   updateFeatureType: (
     featureTypeArgs: FeatureTypeUpdateArgs
   ) => Promise<FeatureType>;
+  // NOTE: The endpoints for dimension type should be used instead of ones for feature and sample type
```
jessica-cheng, line 112: TODO: Remove later

frontend/packages/@depmap/dataset-manager/src/components/index.tsx
```diff
@@ -47,8 +45,20 @@ export default function Datasets() {
     [key: string]: string;
   } | null>(null);
 
+  // TODO: Remove
```
pgm, line 48: Very minor, but it'd be good to include the reason why like you did elsewhere (ie: "The endpoints for feature type and sample type are deprecated and should not be used.")

Seeing a TODO like "Remove" leaves me wondering why.

pgm (approved):
LGTM. Thanks for also flagging all the depreciated methods.

PR #87, opened 2024-10-15
feat(breadbox): SliceQuery support (get dimension data endpoint) 

snwessel:
[Asana](https://app.asana.com/0/1165651979405609/1208505157799193/f)

**This accomplishes two goals:**
* Defining a generalizable endpoint that can replace more feature-specific DE2 endpoints like `/get_metadata`
* Building some functionality (`get_slice_data`) that will be necessary to support DE2/Context Manager context requests. 

**This includes the creation of a new Breadbox endpoint:** Get Dimension Data, `/datasets/dimension/data`
* Parameters: [SliceQuery](https://github.com/broadinstitute/depmap-portal/blob/45058fd5f3a8ca5b17a80e79b9e87c59d378a8dd/depmap-compute/depmap_compute/slice.py#L7) object
* Returns: a dictionary with 3 fields (each a list of equal length):
    * ids
    * labels
    * values


breadbox/breadbox/api/datasets.py
```diff
@@ -456,6 +459,53 @@ def get_dimensions(
     return search_index_entries
 
 
+@router.post(
+    "/dimension/data/",
```
snwessel, line 463: I was on the fence about whether to make this endpoint a `POST` or `GET` method. The reason I chose `POST` here is because I thought it would be nice to keep our requests as close to a [SliceQuery](https://docs.google.com/document/d/175BBPZkNlMuvk_F-xYLs0xDphRZP66JCesUi7RWZECg/edit) JSON object as possible for consistency. 

With a `POST` method, you can use a request body like:
```
{
  "dataset_id": "string",
  "identifier": "string",
  "identifier_type": "feature_id"
}
```

Instead of needing to put the parameters in the URL. Having the request body (with no need for URL encodings), also makes things much more human-readable and therefor easier to debug in the browser (in my mind).

pgm: I agree with this logic. I wish `GET`s could have json bodies, but alas they can't and thus `POST` is the way to go.

breadbox/breadbox/api/datasets.py
```diff
@@ -456,6 +459,53 @@ def get_dimensions(
...
+        identifier_type=identifier_type.name,
+    )
+    slice_values_by_id = slice_crud.get_slice_data(
+        db, settings.filestore_location, parsed_slice_query
+    )
+
+    # Load the labels separately, ensuring they're in the same order as the other values
+    slice_ids: list = slice_values_by_id.index.tolist()
+    labels_by_id = slice_crud.get_labels_for_slice_type(db, parsed_slice_query)
+    slice_labels = [labels_by_id[id] for id in slice_ids] if labels_by_id else None
```
pgm, line 500: I always have a bias away from `Optional[X]` types such as `labels` seems to be. Thinking this through, I wonder under which conditions will `labels` be None?

I thought we had a design principal that all features have a user displayable name (using the id as a default if one wasn't explicitly stated)?

In which case this would be:

```
slice_labels = [labels_by_id[id] for id in slice_ids] if labels_by_id else slice_ids
```

Do you think it's important for this API make a distinction between IDs and labels? (As a consumer of the API, I'd typically want to have what's returned to me be as consistent as possible.)

breadbox/breadbox/crud/dataset.py
```diff
@@ -1391,68 +1393,81 @@ def get_dataset_feature_by_given_id(
...
-            db.query(TabularColumn)
-            .join(TabularCell)
-            .filter(
-                TabularColumn.dataset_id == feature_metadata_dataset_id,
-                TabularColumn.given_id == "label",
-                TabularCell.value == feature_label,
-            )
-            .with_entities(TabularCell.dimension_given_id)
-            .one_or_none()
+    labels_by_given_id = get_dataset_feature_labels_by_id(db, db.user, dataset)
```
pgm, line 1440: Has potential to be a performance issue, but I'm content to leave it until it is. ;)

pgm (approved):
One question about return None for labels, but otherwise looks good to me.

breadbox/breadbox/api/datasets.py
snwessel: Good point! And yeah, you're right, that is the pattern we've been doing elsewhere. I'll update this to match

breadbox/breadbox/crud/dataset.py
snwessel: Yeah I agree. I think in general, the lookups by label have iffy performance right now. 

If we do run into problems loading a lot of dimensions, we can either:
* look them up by ID instead of label (which I think we're trying to move towards anyway) 
* or improve the performance here

breadbox/breadbox/api/datasets.py
snwessel: Actually, as I was updating this, I realized that I don't currently have the behavior we want for column lookups (looking up column data is currently the case that's returning `None` for labels, and that's not what we want). 

When you load a column from a tabular dataset, that column should be indexed by a feature/sample given id. And that feature/sample will have corresponding labels. I'll work on updating that.

snwessel: One last follow-up on this: I realized there are a couple of open questions I have about tabular dataset indexes, and I would like to have a better understanding of how they work before I implement the crud functions we need to return labels for them here. 

I put a discussion topic in our breadbox meeting list: https://app.asana.com/0/1206925945720127/1208554346294686/f

And I put a task in Asana to follow up on this: https://app.asana.com/0/1165651979405609/1208554346294688/f

For now, I'll just do what you had originally suggested and return the IDs as labels for this particular case.

PR #88, opened 2024-10-15
Dimension type form error

jessica-cheng:
Screenshot 2024-10-15 at 5 22 59 PM
Screenshot 2024-10-16 at 9 36 46 AM


frontend/packages/@depmap/dataset-manager/src/components/index.tsx
```diff
@@ -47,8 +45,20 @@ export default function Datasets() {
     [key: string]: string;
   } | null>(null);
 
+  // TODO: Remove
```
pgm, line 48: I think I see somethings from the other PR that I just approved. I'll ignore in this one.

pgm (approved):
LGTM

PR #90, opened 2024-10-16
Fix dataset metadata form

jessica-cheng:
This PR includes the following changes:
- Fixed bug: closing form modal sets state storing dataset to edit as null causing the edit dataset form to not show
- Fixed bug: Making updates to dataset metadata does not update the dataset fields in the UI meaning the UI still showed the old dataset metadata
- Previously the dataset metadata column has button to show a form where you can edit the dataset metadata. Changed it to only view the dataset metadata
- Cleaned up the DatasetMetadata component

Screenshot 2024-10-16 at 5 39 42 PM
Screenshot 2024-10-16 at 5 39 59 PM
Screenshot 2024-10-16 at 5 40 06 PM


frontend/packages/@depmap/dataset-manager/src/components/index.tsx
```diff
@@ -36,14 +35,15 @@ export default function Datasets() {
...
   const [datasetToEdit, setDatasetToEdit] = useState<Dataset | null>(null);
-  const [datasetMetadataToEdit, setDatasetMetadataToEdit] = useState<{
-    [key: string]: string;
-  } | null>(null);
+  const [datasetMetadataToShow, setDatasetMetadataToShow] = useState<
+    | {
+        [key: string]: string;
+      }
+    | null
+    | undefined
```
pgm, line 45: Why the need for two different representations of "nothing"? Is there a distinction between when this is set to `null` and when it's set to `undefined`?

pgm (approved):
Minor things I wondered about, but generally looks fine.

frontend/packages/@depmap/dataset-manager/src/components/DatasetMetadataForm.tsx
```diff
@@ -51,7 +39,7 @@ export default function DatasetMetadataForm({
   );
   const [validValues, setValidValues] = useState<Set<string>>(new Set());
   const [invalidValues, setInvalidValues] = useState<Set<string>>(new Set());
-  const [metadata, setMetdata] = useState({});
+  const [, setMetadata] = useState({});
```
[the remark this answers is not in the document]
jessica-cheng: Good catch! I think during my cleanup I forgot to remove this

frontend/packages/@depmap/dataset-manager/src/components/index.tsx
jessica-cheng: I was trying to adhere to the `Dataset` typescript typing where `dataset_metadata` could be both null and undefined as in not being a property returned in the Dataset response but I think that the property is always returned so I've updated the typing now.

PR #97, opened 2024-10-21
Resources page new accordions

jessica-cheng:
Replaced old Accordion component with React bootstrap's Panel component. This makes it easier to set a default open Accordion/Panel when first entering Resources Page unlike before where I had to add a hack to set state for the old buggy Accordion component

https://github.com/user-attachments/assets/58a57099-d548-4614-b7d2-19060e971526

Reference [old PR for old implementation](https://github.com/broadinstitute/depmap-portal/pull/74)

### DIFF
--- a/frontend/packages/portal-frontend/src/resources/components/ResourcesPage.tsx
+++ b/frontend/packages/portal-frontend/src/resources/components/ResourcesPage.tsx
@@ -1,10 +1,8 @@
 import * as React from "react";
 import { useMemo } from "react";
-import { ListGroup, ListGroupItem } from "react-bootstrap";
-import { Accordion } from "@depmap/interactive";
+import { ListGroup, ListGroupItem, Panel, PanelGroup } from "react-bootstrap";
 import { Link, useLocation } from "react-router-dom";
 import styles from "src/resources/styles/ResourcesPage.scss";
-// import { CollapsiblePanel } from "src/dataPage/components/CollapsiblePanel";
 
 interface ResourcesPageProps {
   subcategories: any;
@@ -23,6 +21,7 @@ export default function ResourcesPage(props: ResourcesPageProps) {
   const { subcategories, defaultTopic } = props;
   console.log(subcategories);
   const query = useQuery();
+  const querySubcategory = query.get("subcategory");
 
   // If window location url has query params at the start, find the post to show
   const initPost = useMemo(() => {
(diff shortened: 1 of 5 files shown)


frontend/packages/portal-frontend/src/resources/components/ResourcesPage.tsx
```diff
@@ -52,43 +53,41 @@ export default function ResourcesPage(props: ResourcesPageProps) {
...
+          {subcategories.map((subcategory: Subcategory) => {
+            let isDefaultExpandedSubcategory: boolean;
+            if (querySubcategory) {
+              isDefaultExpandedSubcategory =
+                subcategory.slug === querySubcategory;
+            } else {
+              isDefaultExpandedSubcategory = !!subcategory.topics.find(
+                (t: Topic) => t.slug === defaultTopic?.slug
+              );
+            }
```
alimourey, line 66: I'm a little confused about what this if/else block is doing. querySubcategory is if a use arrives on this page from a url query param to a specific resource? But if the user doesn't arrive here from a query param, there might also be a default category set to be open on load?

alimourey (commented):
Looks good! I just had a couple of questions.

frontend/packages/portal-frontend/src/resources/components/ResourcesPage.tsx
```diff
@@ -52,43 +53,41 @@ export default function ResourcesPage(props: ResourcesPageProps) {
...
-                        {topic.title}
-                      </ListGroupItem>
-                    </Link>
-                  );
-                })}
-              </ListGroup>
-            </Accordion>
-          );
-        })}
+        <PanelGroup>
```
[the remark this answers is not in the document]
jessica-cheng: That's fair! You're right that I probably don't need PanelGroup. 

I think I initially used it because I noticed it was how they were grouping Panel components. I thought it would be useful to use their toggle props but I believe `activeKey` only toggle opens one accordion at a time while closing the other that doesn't match the `activeKey` (correct me if I'm wrong). I wanted to be able to have more than one panel/accordion open at the same time which is why I made `Panel` into a controlled component instead in `SubcategoryPanel.tsx`

frontend/packages/portal-frontend/src/resources/components/ResourcesPage.tsx
jessica-cheng: Yes exactly! I think maybe the syntax here is confusing for `isDefaultExpandedSubcategory = !!subcategory.topics…` I believe that was the autosuggest from typescript eslint meaning if there is a value found, set to `true` otherwise set to `false`

frontend/packages/portal-frontend/src/resources/components/ResourcesPage.tsx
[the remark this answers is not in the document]
alimourey: I see, that makes sense! You are correct about activeKey.

frontend/packages/portal-frontend/src/resources/components/ResourcesPage.tsx
```diff
@@ -23,16 +23,17 @@ export default function ResourcesPage(props: ResourcesPageProps) {
   const { subcategories, defaultTopic } = props;
   console.log(subcategories);
```
alimourey, line 24: Spotted a console.log. Not sure if you meant to remove this.

alimourey (approved):

PR #99, opened 2024-10-22
feat(breadbox): Define context endpoints on Breadbox

snwessel:
[Asana](https://app.asana.com/0/1165651979405609/1208505020178610/f)

**Background:**
* Randy and I had been working together on designing a "version 2" of the context endpoints that DE2 and context manager use regularly. The goal of this second version was to have a contract we feel happy with long-term, which can also be supported on breadbox. 
* This new API contract has already been implemented in the `portal-backend`. Those existing portal-backend endpoints can combine data from both breadbox and the legacy database (or query whichever source is appropriate). For the time being, I'm assuming those be used by DE2 (not the ones I'm defining here). 
* Meanwhile, we have also been discussing a strategy of re-implementing important DE2 endpoints in breadbox (and putting any shared code in the depmap-compute module). This isn't _really_ necessary for DE2 yet, but is a pre-requisite for Elara (DE2 on Breadbox), and I wanted to try out this approach as early as possible to see if I encounter any unexpected issues with the shared module. 

**The endpoints:** 
Note: these are put under the "temporary" prefix to convey to outside API users that they shouldn't rely on them, not because we're actually planning to change them. More discussion [here](https://app.asana.com/0/1206925945720127/1207990151618883/f).
* `/temporary/context `: Returns all IDs and labels matching the given context
* `/temporary/context/summary`: Returns summary information about how many candidates and matches there were
 
These contracts should almost exactly match the portal-backend implementation. The only differences are:
* When called, these breadbox endpoints _only_ return breadbox data. 
* The URLs are slightly different (ex. `/temporary/context` instead of `/data_explorer_2/v2/context`)
* The breadbox versions have more strict typing, which means better documentation and clearer error messages for badly-formatted requests 🎉 

**Note for reviewers:** Feel free to ignore all of the `breadbox_client` files. They're auto-generated and won't be used for these endpoints.


breadbox/breadbox/schemas/context.py
```diff
@@ -19,4 +24,15 @@ class Context(BaseModel):
...
+
+    # The "dimension_type" field is used by contexts in the new format (with slice queries)
+    # The "context_type" field is used by the old format (with slice IDs)
+    # The two fields contain the same information (the dimension type name)
+    dimension_type: Optional[str] = None
+    context_type: Optional[str] = None
+    vars: dict[str, dict[str, str]] = {}
+
+
+class ContextRequest(BaseModel):
```
jessica-cheng, line 37: Is there a reason why we need to create a new type `ContextRequest`? It seems like ContextRequest is used to type the request parameter but we are ultimately only interested in accessing contents in its one and only field `context`. Why not just type the parameter as `Context` or rename it to `ContextRequest`?

jessica-cheng (commented):
I think I can see the purpose of these endpoints but I wonder if it makes sense to add that as part of Breadbox and am worried about the coupling that seems to be happening between Breadbox and some legacy details from Data Explorer. I don't know if it makes sense to introduce the concept of slice into breadbox. As I'm reading through the new breadbox temp endpoints it ultimately uses some very basic breadbox crud functions/queries. I know @rcreasi has mentioned about creating a layer or maybe interface for Data Explorer specific operations that utilizes Breadbox and I feel like maybe we should reopen that discussion

breadbox/breadbox/schemas/context.py
jessica-cheng: Alternatively, you could also do something like:
```
class ContextRequest(Context):
    pass
```

snwessel: Good question. I agree that this is not a totally normal way to model a request schema. I only added this outer wrapper because I wanted to be able to support the exact same legacy request format we've been using for the past year or so (in addition to the newer, more-flexible, easier-to-read format). 

The reason I'm hesitant to change the format is because these "contexts" are stored in cookies in the users' browsers. I'm not actually 100% sure whether this full object is stored as a cookie, or only the inner dictionary, but regardless I'm hesitant to make Randy do even more refactoring unless there's a clear reason for it. My goal with these new context endpoints was to:
1. Continue to support the old "context" request format
2. Also support a new request format, which is more flexible and supports tabular lookups
3. Return a response which includes more information (both IDs and labels)

For example, a legacy context request is formatted like:
```
{"context": {
    "context_type": "depmap_model",
    "expr": {">": [{"var": f"slice/{dataset_id}/gene1/label"}, 0.05]},
}}
```

Instead of (what I'm guessing you're expecting):
```
{
    "context_type": "depmap_model",
    "expr": {">": [{"var": f"slice/{dataset_id}/gene1/label"}, 0.05]},
}
```
I agree that the cleanest thing to do would be to update it, I'm just not sure if it's worth the effort of refactoring the things that use the old format.

snwessel:
> Yes, let's reopen that discussion! Conveniently, we have the Breadbox Review meeting tomorrow. I think we should add as little to the Breadbox API as possible and to my mind that means dropping support for any legacy behavior.

Oh interesting! Yes, let's reopen the discussion then. If there's a way to move forward without supporting the legacy contexts, I'm all ears! 

And yeah, I think it would be good to have a broader conversation about where feature-specific logic lives long-term (are we moving more things to breadbox?). Phil had added a [similar topic](https://app.asana.com/0/1206925945720127/1208553414820117/f) last week, and I think we got somewhat sidetracked by talking about 3-tiered architectures in breadbox.

breadbox/breadbox/api/temporary.py
```diff
@@ -0,0 +1,81 @@
...
+
+    # Evaluate each of the dimension's given_ids against the context
+    ids_matching_context, _ = slice_crud.get_ids_and_labels_matching_context(
+        db, settings.filestore_location, context
+    )
+
+    return ContextSummary(
+        num_candidates=len(all_labels_by_id.keys()),
+        num_matches=len(ids_matching_context),
+    )
```
jessica-cheng, line 81: These endpoints look very similar and I would advocate for maybe keeping the former while doing the simple counts on the frontend

jessica-cheng (commented):
Okay I'm convinced that the idea of "context evaluation" should live in Breadbox. My initial confusion was that I thought we could simply call `get_subsetted_tabular/matrix_dataset` but after talking some more, it seems like we are actually combining "filtered" slices from multiple datasets together.
I still think that we shouldn't be introducing "legacy slice ids" into breadbox and it sounds like this is something that can be easily addressed according to @rcreasi

breadbox/breadbox/api/temporary.py
snwessel: This morning, @rcreasi and I talked about this and it seems like he's on-board with combining them as long as it's still possible to get the number of "candidate" entities. I would propose that I:
1. get rid of this "summary" endpoint altogether
2. add an additional field `"num_candidates"` to the response from the other context endpoint (since it's not easy to get this information through any other breadbox endpoints) 

Does that sound reasonable to you, @jessica-cheng ?

snwessel:
@jessica-cheng I believe I've implemented all your suggestions now, including:
1. Dropping support for the legacy context request format
2. Combining the two endpoints into one
3. Removing the `ContextRequest` type, which was wrapper around the context requests

And it sounds like this is blocking Randy, so I am going to merge the branch once all the tests have passed 👍 

If there was any other feedback you had that I missed, feel free to let me know.

breadbox/breadbox/api/temporary.py
jessica-cheng: Sounds good!

breadbox/breadbox/crud/dataset.py
```diff
@@ -1029,6 +1029,36 @@ def get_dataset_sample_labels_by_id(
...
 
+def get_dimension_labels_by_id(
+    db: SessionWithUser, dimension_type_name: str
+) -> dict[str, str]:
+    """
+    For a given dimension, get all IDs and labels that exist in the metadata.
+    """
+    dimension_type: Optional[DimensionType] = db.query(DimensionType).filter(
+        DimensionType.name == dimension_type_name
+    ).one_or_none()
```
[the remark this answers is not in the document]
snwessel: Oh thank you, that's a good suggestion. I just fixed it by moving this whole `get_dimension_labels_by_id` function into `crud/types.py`, so it can access `get_dimension_type`.  

While I was working on that, I also found about 4 other paces that _could_ be using this same `get_dimension_type` function instead re-implementing the query. However, I can't easily update them without creating a circular import (by importing `crud/types.py` into `crud/dataset.py`).

This is one of the reasons I would really like to make these crud files smaller and move a few things like label-loading into a service layer.

PR #110, opened 2024-10-29
update Elara ContextBuilder to use /temp/context endpoint

rcreasi:


frontend/packages/elara-frontend/src/pages/DataExplorer/api.ts
```diff
@@ -46,6 +48,33 @@ const postJson = async <T>(url: string, obj: unknown): Promise<T> => {
...
+  const varsAsSliceQueries = Object.fromEntries(
+    Object.entries(context.vars).map(([varName, variable]) => [
+      varName,
+      {
+        // The `DataExplorerContextVariable` type has some extra fields that
+        // aren't part of the SliceQuery format. We'll strip those out so
+        // the backend doesn't get confused.
+        dataset_id: variable.dataset_id,
+        identifier: variable.identifier,
+        identifier_type: variable.identifier_type,
```
rcreasi, line 61: I figured out this is why the endpoint was always returning zero matches.

In the `ContextBuilder`, I needed to keep track of extra information about each variable.
https://github.com/broadinstitute/depmap-portal/blob/dff1a53f1a9c79459664e794366650292aaa335f/frontend/packages/@depmap/types/src/data-explorer-2.ts#L11-L15

I was serializing all of it, figuring the backend would ignore the extra fields. But it doesn't. This line silently fails:

https://github.com/broadinstitute/depmap-portal/blob/dff1a53f1a9c79459664e794366650292aaa335f/depmap-compute/depmap_compute/context.py#L124

I get that it doesn't want to construct a `SliceQuery` from the dict with extra keys but it doesn't throw an error. It just keeps running with bad results.

frontend/packages/@depmap/data-explorer-2/src/components/ContextBuilderV2/hooks/useNumMatches.ts
```diff
@@ -50,14 +43,15 @@ function useNumMatches(expr: Expr) {
 
       (async () => {
         try {
-          const summary = await fetchContextSummaryV2({
+          const result = await api.evaluateContext({
+            name: "",
```
snwessel, line 47: I wasn't sure if this context `name` field was actually being used anymore so I made it optional on the breadbox endpoint. If you're not using it anymore (and we're not supporting legacy contexts in breadbox), maybe I should remove it altogether?

Also, breadbox should be fine with `name` being unset, so you could just exclude it from your request here if it's not doing anything.

snwessel (approved):
Looks great to me!

frontend/packages/elara-frontend/src/pages/DataExplorer/api.ts
rcreasi: I forget the exact scenario but there was situation where one unknown feature that was causing the whole thing to fail. But if that comes up again, we should handle it a much more targeted way. So yeah, you can ditch that.

That's not where the error was getting swallowed, though. I did some digging into the JsonLogic library and it's designed to use a default value (or `None`) if the dictionary lookup throws an error 🙁

https://github.com/nadirizr/json-logic-py/blob/5fda9125eab4178f8f81c7779291940e31e87bab/json_logic/__init__.py#L109-L110

frontend/packages/@depmap/data-explorer-2/src/components/ContextBuilderV2/hooks/useNumMatches.ts
rcreasi: There are a few operations where the name matters (some plotting stuff that we may or may not wind up implementing in Breadbox) but this certainly isn't one of them. So we could make this endpoint agnostic to it.

frontend/packages/elara-frontend/src/pages/DataExplorer/api.ts
snwessel: Oh interesting! 🤔

rcreasi: We could re-raise it as `LookupError`. It wouldn't swallow that.

```python
if var_name not in self.cache:
    try:
        slice_query = SliceQuery(**self.slice_query_vars[var_name])
        self.cache[var_name] = self.get_slice_data(slice_query).to_dict()
    except (KeyError, TypeError, ValueError) as e:
        raise LookupError from e
```

PR #112, opened 2024-10-30
Deprecate post sample feature type

jessica-cheng:
- Remove deprecated SampleType and FeatureType methods from dataset manager and replace with DimensionType methods
- Remove old SampleType and FeatureType type references and replace with new DimensionType types. 

**NOTE**: _Only Elara uses the deprecated methods and types now since removing them will be more effort to not upset the type checker/lint errors._


pgm (approved):
Sounds good to me.

PR #113, opened 2024-10-30
feat(breadbox): Support tabular dataset metadata lookups

snwessel:
[Asana](https://app.asana.com/0/1165651979405609/1208554346294688/f)

**The problem:** For tabular datasets, there weren't any reusable methods for loading the labels (or other metadata) associated with a tabular dataset's index. As a result, when when the `/temp/context/` or `/datasets/dimension/data/` endpoints are called for a "column" slice type, they currently just return IDs as the labels (instead of getting the actual labels).

**The solution:** Defining a few new reusable methods:
* `get_dimension_type_metadata_col` - Gets a column of values from the given dimension type's metadata. 
* `get_tabular_dataset_index_given_ids` - identify all of the row IDs belonging to a tabular dataset. This can be used to join the dataset's metadata. 
* `get_tabular_dataset_metadata_annotations` - Gets a column of values from the given tabular dataset's metadata (this is a subset of the full dimension type's metadata) 
* `get_tabular_dataset_labels_by_id` Get all of the labels for the index of a tabular dataset. If a tabular dataset does not have any metadata associated with it, use the given IDs as labels. 

**Setting up a service layer:** I also wanted to use this as an opportunity to create a "service" module/layer, as we'd discussed at our last breadbox meeting (notes [here](https://app.asana.com/0/0/1208553414820117/1208578469618061/f)). The reason I'm setting this up now is that I couldn't define a function like `get_tabular_dataset_labels_by_id` in the dataset crud without creating a circular import (because it needs to import methods from `crud/types.py,` which in turn imports from `crud/dataset.py`). Instead of dealing with the circular imports, I thought it might be time to make a "service" directory. 

**Future steps:** There are 3-4 other places that re-implement a query very similar to  `get_dimension_type_metadata_col` (at least [here](https://github.com/broadinstitute/depmap-portal/blob/0820da05015e65ac76fceb2d04a04ba734aa9df1/breadbox/breadbox/crud/dataset.py#L1061), [here](https://github.com/broadinstitute/depmap-portal/blob/0820da05015e65ac76fceb2d04a04ba734aa9df1/breadbox/breadbox/crud/dataset.py#L1109), and [here](https://github.com/broadinstitute/depmap-portal/blob/0820da05015e65ac76fceb2d04a04ba734aa9df1/breadbox/breadbox/crud/dataset.py#L1527)). I would eventually like to combine these and remove the duplication.

### DIFF
--- a/breadbox/breadbox/models/dataset.py
+++ b/breadbox/breadbox/models/dataset.py
@@ -116,7 +116,6 @@ class TabularDataset(Dataset):
     index_type_name = Column(
         String, ForeignKey("dimension_type.name", ondelete="CASCADE"), nullable=False
     )
-    index_type = relationship("DimensionType", foreign_keys=[index_type_name])
     __mapper_args__ = {"polymorphic_identity": "tabular_dataset"}
 
     @property
--- a/breadbox/breadbox/models/dataset.py
+++ b/breadbox/breadbox/models/dataset.py
@@ -116,6 +116,7 @@ class TabularDataset(Dataset):
     index_type_name = Column(
         String, ForeignKey("dimension_type.name", ondelete="CASCADE"), nullable=False
     )
+    index_type = relationship("DimensionType", foreign_keys=[index_type_name])
     __mapper_args__ = {"polymorphic_identity": "tabular_dataset"}
 
     @property
--- a/breadbox/breadbox/service/metadata.py
+++ b/breadbox/breadbox/service/metadata.py
@@ -1,0 +1,118 @@
+import logging
+from typing import Any, Optional
+
+from breadbox.crud import dataset as dataset_crud
+from breadbox.crud import types as types_crud
+from breadbox.db.session import SessionWithUser
+from breadbox.schemas.custom_http_exception import ResourceNotFoundError
+from breadbox.models.dataset import (
+    Dataset,
+    MatrixDataset,
+)
+
+from depmap_compute.slice import SliceQuery
+
+log = logging.getLogger(__name__)
+
+
+def get_matrix_dataset_feature_annotations(
+    db: SessionWithUser, user: str, dataset: MatrixDataset, metadata_col_name: str,
+) -> dict[str, Any]:
+    """
+    For the given dataset, load metadata of the specified type, keyed by given id.
+    If there is no metadata of this type, return an empty dictionary.
+    """
+    if dataset.feature_type is None or dataset.feature_type.dataset_id is None:
+        return {}
+
+    full_metadata_col = types_crud.get_dimension_type_metadata_col(
+        db, dimension_type_name=dataset.feature_type, col_name=metadata_col_name
+    )
+    # Filter the metadata to only include the given IDs belonging to this dataset
+    dataset_features = dataset_crud.get_dataset_features(db, dataset, user=db.user)
+    dataset_given_ids = [feature.given_id for feature in dataset_features]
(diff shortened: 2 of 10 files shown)


breadbox/breadbox/service/metadata.py
```diff
@@ -0,0 +1,71 @@
...
+    If there are no labels in the metadata or there is no metadata, then just return the given IDs as labels.
+    """
+    metadata_labels = get_tabular_dataset_metadata_annotations(
+        db=db, dataset=dataset, metadata_col_name="label"
+    )
+    if metadata_labels:
+        return metadata_labels
+    else:
+        columns = dataset_crud.get_dataset_columns(db=db, dataset=dataset)
+        return {column.given_id: column.given_id for column in columns}
```
snwessel, line 47: Oh actually, after I opened this PR, I realized that this particular case is going to cause issues because it's loading the given IDs by column, not by row. I'll work on fixing this and then let you know when it's actually ready for review. Sorry about that!

snwessel: Okay sorry for the confusion, this should actually be ready to review now. 

@jessica-cheng I added you as a reviewer because I think you're the most familiar with the existing tabular dataset stuff (I'm relatively new to this part of breadbox and was learning my way around the existing crud functions we have). Any feedback is welcome!

breadbox/breadbox/models/dataset.py
```diff
@@ -116,6 +116,7 @@ class TabularDataset(Dataset):
     index_type_name = Column(
         String, ForeignKey("dimension_type.name", ondelete="CASCADE"), nullable=False
     )
+    index_type = relationship("DimensionType", foreign_keys=[index_type_name])
```
snwessel, line 119: I added this relationship field so it matches the `sample_type` and `feature_type` fields we have on the matrix datasets.

breadbox/breadbox/service/metadata.py
```diff
@@ -0,0 +1,80 @@
...
+    For the given tabular dataset, load a column from the associated metadata.
+    Return a dictionary of values keyed by given id.
+    Note: this should work regardless of whether or not the given dataset IS metadata
+    or HAS metadata.
+    """
+    full_metadata_col = types_crud.get_dimension_type_metadata_col(
+        db, dimension_type_name=dataset.index_type_name, col_name=metadata_col_name
+    )
+    # Filter the metadata to only include the given IDs belonging to this dataset
+    dataset_index_given_ids = dataset_crud.get_tabular_dataset_index_given_ids(
```
jessica-cheng, line 28: potentially redundant if tabular dataset is already a dimension type metadata but I think it's fine

breadbox/breadbox/service/metadata.py
```diff
@@ -0,0 +1,71 @@
...
+from breadbox.schemas.custom_http_exception import ResourceNotFoundError
+from breadbox.models.dataset import MatrixDataset, TabularDataset
+
+from depmap_compute.slice import SliceQuery
+
+log = logging.getLogger(__name__)
+
+
+def get_tabular_dataset_metadata_annotations(
+    db: SessionWithUser, dataset: MatrixDataset, metadata_col_name: str,
```
jessica-cheng: The typing here for dataset is `MatrixDataset` but the name says `get_tabular_dataset_metadata_annotations` that is confusing and makes me wonder if your tests are correct

jessica-cheng: Oh weird it looks like I accidentally reviewed the outdated version of this file. Disregard the above comment

PR #114, opened 2024-10-30
fix(breadbox): Improve error handling for context requests

snwessel:
Currently, when the `ContextEvaluator` runs into a problem, it just hides the errors returns 0 matches. 

@rcreasi had made some helpful suggestions ([here](https://github.com/broadinstitute/depmap-portal/pull/110#discussion_r1821452610)) for how error handling could be improved. I implemented those changes and added a couple of tests to make sure it's now actually returning reasonable errors to users.


rcreasi (approved):
Yay! Error handling for the win :100:

I'm glad we caught this now. If `get_data_slice()` fails (and I'm sure someday it will), the error would've been similarly swallowed and that would've been pretty bad 😅

PR #117, opened 2024-11-04
feat(breadbox): Only log to error reporter if status code is a 500

pgm:
Tiny change to stop sending non-500 errors to error reporter


snwessel (approved):
Looks good to me!

PR #122, opened 2024-11-05
fix(breadbox): workaround for memory problem reading parquet files.

pgm:
Very small change to work around issue with pd.read_parquet(). (memory utilization exploding when trying to read a large-ish parquet file)

### DIFF
--- a/breadbox/breadbox/io/data_validation.py
+++ b/breadbox/breadbox/io/data_validation.py
@@ -177,7 +177,7 @@ def _validate_data_value_type(
         return df.astype(np.float64)
 
 
-def _read_parquet(file: BinaryIO, value_type: ValueType) -> pd.DataFrame:
+def _read_parquet(file, value_type: ValueType) -> pd.DataFrame:
     # It appears that pd.read_parquet() by default uses pyarrow. However, for some reason
     #  when reading a file with 20k columns, the memory usage balloons
     # to > 30GB and would take down breadbox. However, using fastparquet seems to avoid this problem.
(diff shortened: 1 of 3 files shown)


breadbox/breadbox/io/data_validation.py
```diff
@@ -83,6 +84,20 @@ def annotation_type_to_pandas_column_type(annotation_type: AnnotationType):
     return dtype
 
 
+def _read_parquet_with_less_memory(filename):
+    # it seems like pd.read_parquet() should be the right thing to do
+    # but in practice, when reading a file with 20k columns, the memory usage balloons
+    # to > 30GB and would take down breadbox. Reading it in as a table, and then
+    # processing it column by column seems to avoid this problem. Not 100% sure that this
+    # results in identical behavior, but it seems valid for numerical matrices
+    table = pq.read_table(filename)
```
snwessel, line 93: If this fixes the issue, I think it looks good and I feel good about you merging it (especially since it's needed for the release).

But I am curious... I found this [github issue](https://github.com/pandas-dev/pandas/issues/55296), suggesting the problem is the pyarrow engine that pandas uses. I'd be curious if you could use a simpler solution of just doing:
```
pandas.read_parquet(engine='fastparquet')
```

snwessel (approved):
Looks good to me! I had one question/comment, but we can revisit it later if it's helpful to just merge this in now

breadbox/breadbox/io/data_validation.py
pgm: The memory leak described there looks fairly different then how I'd characterize what I've observed. However, that's a good point about maybe the issue stems from pyarrow and not pandas.

I tried
```
pandas.read_parquet(engine='fastparquet')
```

and yes, that works fine. We'll have to add fastparquet as a dependency, but that sounds fine.

I'll switch the code to use that instead.

PR #124, opened 2024-11-06
Changed settings so that we can specify multiple directories containing downloads configs

pgm:
Previously the portal would check for downloads looking at two directories based on the DOWNLOADS_PATH and SHARED_DOWNLOADS_PATH config parameters. We were putting the environment specific downloads at DOWNLOADS_PATH and the common downloads into SHARED_DOWNLOADS_PATH. However, we increasingly have environments which we want to share more downloads.

This change allows us to generalize from checking two directories to checking an arbitrary list of directories.


snwessel (approved):
Looks good to me. It'll be nice to have these combined! Although I noticed a couple other files that might need to be updated as well:
* `portal-backend/overrides.py`
* `config/dev/settings.py `

since they seem to reference the same `DOWNLOADS_PATH` and `SHARED_DOWNLOADS_PATH` variables

pgm:
> Looks good to me. It'll be nice to have these combined! Although I noticed a couple other files that might need to be updated as well:
> 
> * `portal-backend/overrides.py`
> * `config/dev/settings.py `
> 
> since they seem to reference the same `DOWNLOADS_PATH` and `SHARED_DOWNLOADS_PATH` variables

Good catch. I've fixed those two as well.

PR #126, opened 2024-11-12
refactor(breadbox): Part 1 of removing duplicate metadata lookups

snwessel:
[Asana](https://app.asana.com/0/1165651979405609/1208651997719944/f)

**I am breaking these changes into two parts so I can more quickly unblock Jessica. This PR includes:**
1. Redefining a couple of methods to remove duplicate SQLAlchemy queries: 
    * For example, `crud.dataset.get_dataset_sample_annotations` (which has a complicated SQLAlchemy query) was replaced by `service.metadata.get_matrix_dataset_sample_metadata` (which calls existing crud functions instead of re-implementing queries) 
3. Moving some methods to the service layer (both because they belong there and because I was getting circular import errors from keeping them in the crud) 
    * `get_dataset_feature_labels_by_id`
    * `get_dataset_sample_labels_by_id`
    * `test_get_dataset_feature_by_label`
    * `test_get_dataset_sample_by_label`
    * `get_subsetted_matrix_dataset_df`

**Part 2** will be replacing some other duplicate crud functions with calls to these new more-reusable methods (I'm guessing this second part shouldn't impact any of Jessica's work)

### DIFF
--- a/breadbox/breadbox/service/metadata.py
+++ b/breadbox/breadbox/service/metadata.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any, Optional
+from typing import Any
 
 from breadbox.crud import dataset as dataset_crud
 from breadbox.crud import types as types_crud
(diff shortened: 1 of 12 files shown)


breadbox/breadbox/service/metadata.py
```diff
@@ -36,6 +42,82 @@ def get_tabular_dataset_metadata_annotations(
...
+        )
+        if metadata_labels_by_given_id:
+            return metadata_labels_by_given_id
+
+    # If there are no labels or there is no feature type, return the given IDs
+    all_dataset_features = dataset_crud.get_dataset_features(db, dataset)
+    return {feature.given_id: feature.given_id for feature in all_dataset_features}
+
+
+def get_dataset_sample_labels_by_id(
```
jessica-cheng, line 104: I like how we are being explicit with what type of dataset in this method's parameter. However, it would also be nice for the name of the method to be more explicit too like `get_matrix_dataset_sample_labels_by_id` similar to the below method `get_tabular_dataset_labels_by_id`

breadbox/breadbox/service/metadata.py
```diff
@@ -36,6 +42,82 @@ def get_tabular_dataset_metadata_annotations(
...
+) -> dict[str, Any]:
+    """
+    For the given matrix dataset, load a column from the associated metadata.
+    The result will only include given ids which exist in both the dataset and the metadata.
+    """
+    full_metadata_col = types_crud.get_dimension_type_metadata_col(
+        db, dimension_type_name=dataset.feature_type_name, col_name=metadata_col_name
+    )
+    # Filter the metadata to only include the given IDs belonging to this dataset
+    dataset_features = dataset_crud.get_dataset_features(db, dataset)
```
jessica-cheng, line 56: Maybe we should rename `get_dataset_features` to `get_matrix_dataset_features`?

jessica-cheng (approved):
Some minor comments about naming but generally LGTM!

PR #127, opened 2024-11-14
feat(breadbox): Added /tmp/cas endpoints for content addressable storage

pgm:
Added content addressable storage implementation which is near identical to the implementation in portal-backend. 
The primary difference is that it stores values in a sqlite dict instead of cloud storage to keep breadbox decoupled from
google cloud services.


snwessel (approved):
The changes look good to me, although I'm not sure I fully understand what this will be used for. 

I see you have a ticket in Asana [here](https://app.asana.com/0/1165651979405609/1208741693810950/f) saying it's blocking the ContextManager updates. 

Is that because this is how we store saved contexts? My current understanding is that the users' contexts are vaguely "stored as cookies", but maybe that's not the full picture of what's going on.

pgm:
@snwessel The more I think about it, the more the reason why we need this is a little hazy. 

I think the original issue may have been that we wanted the ability to generate URLs for plots in data explorer, but that includes potentially encoding the context into the URL. And contexts can potentially be more than a few kbs and exceed the cap on how long a URL can be.

So, we introduced CAS as a way to pass contexts around with much less space. Instead of passing the context as value, we pass a CAS address, which can be resolved to a value.

Regardless of the original intentions, DE2 now has a dependency that it can put/get contexts from CAS, and so it was a blocking issue not having this capability in BB.

PR #129, opened 2024-11-14
feat(breadbox): Get dimension identifiers

jessica-cheng:
New endpoint as outlined in [proposal](https://docs.google.com/document/d/1vskDzpcl37hFELp-_fh8Af9PiJjeJO6tleK9cLQb5PM/edit?pli=1&tab=t.0)

@snwessel I considered using your new metadata lookup methods but decided against it bc I figured the performance for looking up each individual dataset would be far worse than a method that looks up the label metadata for multiple datasets in a single query

NOTE: This implementation makes changes to the crud function `get_datasets`. This appeared to have broken an existing test however and I'm looking into it
TBD: Write more comprehensive test for new endpoint

### DIFF
--- a/breadbox/breadbox/api/types.py
+++ b/breadbox/breadbox/api/types.py
@@ -603,21 +603,16 @@ def list_dimension_types_endpoint(db: SessionWithUser = Depends(get_db_with_user
 def get_dimension_type_identifiers(
     name: str,
     data_type: Annotated[Union[str, None], Query()] = None,
-    show_only_dimensions_in_datasets: Annotated[Union[bool, None], Query()] = None,
+    show_only_dimensions_in_datasets: Annotated[bool, Query()] = False,
     db: SessionWithUser = Depends(get_db_with_user),
 ):
     dim_type = type_crud.get_dimension_type(db, name)
     if dim_type is None:
         raise HTTPException(404, f"Dimension type {name} not found")
 
-    if show_only_dimensions_in_datasets is True:
-        dimension_ids_and_labels = metadata_service.get_dimension_type_identifiers(
-            db, dim_type, data_type, show_only_dimensions_in_datasets=True
-        )
-    else:
-        dimension_ids_and_labels = metadata_service.get_dimension_type_identifiers(
-            db, dim_type, data_type
-        )
+    dimension_ids_and_labels = metadata_service.get_dimension_type_identifiers(
+        db, dim_type, data_type, show_only_dimensions_in_datasets
+    )
 
     return [
         DimensionIdentifiers(id=id, label=label)
--- a/breadbox/breadbox/api/types.py
+++ b/breadbox/breadbox/api/types.py
@@ -24,7 +24,12 @@ from ..io.data_validation import validate_dimension_type_metadata
 from ..schemas.custom_http_exception import UserError, ResourceNotFoundError
 from breadbox.models.dataset import AnnotationType
 import breadbox.crud.dataset as dataset_crud
-from breadbox.schemas.types import DimensionType, UpdateDimensionType, AddDimensionType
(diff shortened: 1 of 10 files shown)


snwessel (approved):
Looks great! 

>  I figured the performance for looking up each individual dataset would be far worse

Yeah I definitely agree. We'll be adding a lot more datasets so it doesn't hurt to optimize a bit now.

breadbox/breadbox/api/types.py
```diff
@@ -580,6 +595,36 @@ def list_dimension_types_endpoint(db: SessionWithUser = Depends(get_db_with_user
...
+        raise HTTPException(404, f"Dimension type {name} not found")
+
+    if show_only_dimensions_in_datasets is True:
+        dimension_ids_and_labels = metadata_service.get_dimension_type_identifiers(
+            db, dim_type, data_type, show_only_dimensions_in_datasets=True
+        )
+    else:
+        dimension_ids_and_labels = metadata_service.get_dimension_type_identifiers(
+            db, dim_type, data_type
+        )
```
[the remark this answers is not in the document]
snwessel: Oh I see now, I hadn't noticed that you were setting `Union[bool, None]` as the typing for that `show_only_dimensions_in_datasets` parameter.

But I am curious why that parameter needs to be `Union[bool, None]` instead of just a `bool`? Couldn't you update the typing to be something like:
```
show_only_dimensions_in_datasets: Annotated[bool, Query()] = False,
```
I think that would make it more clear what the default behavior is. Also, I just tried it out locally and it seems to work fine at least with the breadbox client (screenshot below), but I'd be curious if you've run into different issues?

 
image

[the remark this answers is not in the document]
jessica-cheng: Ah maybe I misinterpreted the FastAPI docs. I think you're right. It makes sense that query params should already be optional without needing to explicitly pass the default value `=None`

breadbox/breadbox/api/types.py
```diff
@@ -644,7 +689,7 @@ def delete_dimension_type_endpoint(
 
     # don't count datasets which are actually the metadata for this type
     datasets_with_using_type = [
-        x for x in datasets_with_using_type if x.id == dim_feature_type.dataset_id
+        x for x in datasets_with_using_type if x.id != dim_feature_type.dataset_id
```
jessica-cheng, line 692: @pgm I encountered a bug which seemed related to this line while testing my new endpoint. I believe based on your comment this change should be your intended behavior. I've also added more test cases to your `test_all_dimension_type_methods` to cover what I believe this line was intending

breadbox/breadbox/crud/dataset.py
```diff
@@ -108,14 +109,28 @@ def get_datasets(
...
+            filter_clauses.append(
+                or_(
+                    dataset_poly.MatrixDataset.feature_type_name == feature_type,
+                    dataset_poly.TabularDataset.index_type_name == feature_type,
+                )
+            )
+        else:
+            filter_clauses.append(
+                dataset_poly.MatrixDataset.feature_type_name == feature_type
+            )
```
jessica-cheng, line 133: This filter clause has been modified so that a dimension type with 'sample' as the axis doesn't get used as a feature_type filter

jessica-cheng (commented):
@snwessel I noticed you already approved the PR. As I was fixing the broken test and testing cases for the new endpoint there has been some changes. Mostly regarding the new filters in `crud/dataset.py` and tests in `tests/api/test_types.py`

breadbox/breadbox/api/types.py
pgm: Yes, you're correct. This change fixes my mistake.

PR #137, opened 2024-11-21
allow contexts to be persisted in Elara

rcreasi:
This updates the Elara version of Context Manager to save contexts to Breadbox. It takes advantage of the new `/temp/cas` and `/temp/cas/{key}` endpoints Phil added in PR #127

The list of contexts for each user is still saved in local storage.


snwessel (approved):
Looks great to me! Thanks for the walk-through this morning!

PR #142, opened 2024-11-27
Add dataset loading

jessica-cheng:
- Add task polling to dataset form on submission
- Add submission message to notify whether dataset upload is successful or failed
- Update dataset table to show newly added dataset

### Example of Loading/submission message in form
https://github.com/user-attachments/assets/ec3ae513-743d-4a16-b797-9620806ca5b5

### Example of caught error
_NOTE: This specific example shows that a column name in the `column_metadata` does not match the actual dataset's column. However, I realize the error message returned isn't very informative so making a more informative error message from the backend may be needed_
Screenshot 2024-11-27 at 12 57 25 PM


pgm (approved):
LGTM

PR #144, opened 2024-12-05
chore(breadbox): Celery checker

jessica-cheng:
- Add util function for celery to check if celery is running
- For all functions that call a celery task, call that util function to check if celery is running first. The goal is to fail fast if celery isn't running
- Added a health check endpoint to check if celery is running (TBD if should keep because the `health_check/ok` endpoint is already calling that utils function)

To be done: Currently tests failing with celery. Need to modify behavior


breadbox/breadbox/api/downloads.py
```diff
@@ -67,7 +69,10 @@ def export_dataset(
     user: str = Depends(get_user),
     settings: Settings = Depends(get_settings),
 ):
-
+    try:
```
pgm, line 72: I wonder if we could consolidate these calls.

Instead of calling utils.check_celery() before job submission, what do you think about putting it in get_task_status? 

That's when the worker being offline is a problem: the status will be reported as pending forever.

breadbox/breadbox/celery_task/utils.py
```diff
@@ -204,3 +204,25 @@ def update_state(
...
     task.update_state(state=state, meta=meta)
+
+
+def check_celery():
+    """
+    Checks to see if celery redis broker is connected.
+    Pings to see if any worker responds and returns true if successful response detected
+    """
+    inspect = app.control.inspect()
+    try:
```
pgm, line 215: I don't know that we need to explicitly connect to the broker before submitting a task. If the broker is offline, then we already get an exception thrown when it tries to submit the task to the broker. 

That being said, I suppose it's harmless to try. Probably just costs a few milliseconds and I can't imagine it having any real downside.

breadbox/breadbox/celery_task/utils.py
```diff
@@ -204,3 +204,25 @@ def update_state(
...
+        # Tries to connect to celery broker
+        conn = app.broker_connection().ensure_connection(max_retries=3)
+    except Exception as exc:
+        raise CeleryConnectionError(
+            "Failed to connect to celery redis broker!"
+        ) from exc
+    # Pings workers to see if any of them respond. Returns None if no response
+    ping = inspect.ping()
```
pgm, line 223: That's interesting... The worker is asynchronous but this call is clearly synchronous. There must be a timeout baked into this. 

Regardless if you've verified that this is reliably identifying when the worker is offline, then looks good to me.

pgm: I looked up `inspect.ping()` and from what I see in this [SO post,](https://stackoverflow.com/questions/8506914/detect-whether-celery-is-available-running) I think any time that there's a backlog of work, will result in an exception being thrown.

Imagine we have 1 worker, and two users. User A submits a "custom analysis" job and while that's running user B submits a batch download. I think user B will get an exception saying "Celery workers are not responding" instead of A's job running and then B's job running.

Is that behavior intentional? Perhaps it's not bad, but I think at the very least, we'd want a longer timeout then the default of 1 second.

Rejecting requests when there's a backlog has some nice properties ("load shedding") but it also has the downside that when the server is busy, instead of jobs taking longer, users will have their job fail.

pgm: I could be wrong though. It's something I think would good to be explicitly try to make sure we understand what's going to happen before deploying this change.

breadbox/breadbox/api/downloads.py
jessica-cheng: Interesting! I noticed half of the functions using celery returned either `get_task_status()` or `cast_celery_task()` so I wasn't sure if I should have put it in `get_task_status` since not all functions calling a celery task calls it. Ideally, I would put them in celery's `delay()` function but patching that function felt too hacky…

I figured the fail fast approach would be better. I also considered creating a decorator or using fastapi's dependency injection `Depends()` but that would still modify each endpoint regardless

breadbox/breadbox/celery_task/utils.py
jessica-cheng: @pgm  Usually if celery isn't running (at least on development) we've forgotten to start the celery broker. I figured returning an explicit error to the client could be helpful.. This does have 3 retries to connect to the broker which takes a few seconds rather than a few milliseconds though so maybe I should decrease the amount of retries

breadbox/breadbox/celery_task/utils.py
jessica-cheng: @pgm yes `ping()` has a default timeout of 1 second. HOWEVER, I just saw a concerning [post](https://github.com/celery/celery/discussions/7240) where calling `ping()` may not return any response if all workers are busy running tasks. I'm thinking the `stats()` like the [top stackoverflow post](https://stackoverflow.com/questions/8506914/detect-whether-celery-is-available-running) suggested would be better

jessica-cheng:
> Looks fine but I personally think the try/catches repeated are unnecessary and are distracting:
> 
> ```
> try:
>         check_celery()
> except CeleryConnectionError as err:
>         raise err
> ```
> 
> What's the motivation for those?

I can get rid of them! I think this was before I decided to put the try/excepts in `check_celery()` instead

breadbox/breadbox/celery_task/utils.py
jessica-cheng: @pgm Oops I didn't see your subsequent comments and you've basically addressed everything I was pointing out including the SO post 😅 You're right, the behavior of throwing an error when workers are occupied wasn't my intention. I think using `stats()` like the top SO reply might be what I want since it seems to return None if there are no workers present. I will test this first before making the change official

pgm (approved):

PR #145, opened 2024-12-10
fix(breadbox): Fix categorical datasets validation

jessica-cheng:
### Fixes:

- Validation for `allowed_values` field incorrect and broken since update to pydantic 2
- Add missing tests for dataset uploads of categorical datasets. Previous categorical dataset tests were for the old POST dataset endpoint
- Clean up some methods

### DIFF
--- a/breadbox/breadbox/schemas/dataset.py
+++ b/breadbox/breadbox/schemas/dataset.py
@@ -180,10 +180,12 @@ class MatrixDatasetParams(SharedDatasetParams):
 
     @field_validator("allowed_values", mode="after")
     @classmethod
-    def check_valid_allowed_values(cls, v: List[str]):
+    def check_valid_allowed_values(cls, v: Optional[List[str]]):
         """
         Checks there are no empty strings and no repeated allowed values. Values in allowed values list are not case-sensitive.
         """
+        if v is None:
+            return v
         # Decision to make allowed values not case-sensitive in case user error in accidental repeats
         allowed_values_list_lower = [str(x).lower() for x in v]
         allowed_values_set = set(allowed_values_list_lower)
--- a/breadbox/breadbox/schemas/dataset.py
+++ b/breadbox/breadbox/schemas/dataset.py
@@ -115,20 +115,6 @@ class SharedDatasetParams(BaseModel):
             raise ValueError("Must be hex string")
 
 
-def check_allowed_values_not_empty(v):
-    if v == "":
-        raise UserError("Empty strings are not allowed.")
-    allowed_values_list_lower = [str(x).lower() for x in v]
-    if len(set(allowed_values_list_lower)) != len(v):
-        raise UserError(
-            msg="Make sure there are no repeats in allowed_values. Values are not considered case-sensitive",
-        )
-    return v
-
-
-AllowedValue = Annotated[str, AfterValidator(check_allowed_values_not_empty)]
-
-
 class MatrixDatasetParams(SharedDatasetParams):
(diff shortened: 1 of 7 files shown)


breadbox/breadbox/schemas/dataset.py
```diff
@@ -192,6 +178,27 @@ def check_valid_value_type_with_allowed_values(self):
...
 
+    @field_validator("allowed_values", mode="after")
+    @classmethod
+    def check_valid_allowed_values(cls, v: Optional[List[str]]):
+        """
+        Checks there are no empty strings and no repeated allowed values. Values in allowed values list are not case-sensitive.
+        """
+        if v is None:
+            return v
+        # Decision to make allowed values not case-sensitive in case user error in accidental repeats
```
jessica-cheng, line 189: This is carried over from our past decisions but I've written the comment to make it more visible. Example of `"true"` and `True` should be considered the same value

breadbox/breadbox/schemas/dataset.py
```diff
@@ -192,6 +178,27 @@ def check_valid_value_type_with_allowed_values(self):
...
+        allowed_values_list_lower = [str(x).lower() for x in v]
+        allowed_values_set = set(allowed_values_list_lower)
+        print(allowed_values_set, allowed_values_list_lower)
```
pgm, line 192: probably should pull out the print stmt before merging in

breadbox/breadbox/schemas/dataset.py
```diff
@@ -192,6 +178,27 @@ def check_valid_value_type_with_allowed_values(self):
...
+        if len(allowed_values_set) != len(v):
+            raise UserError(
+                msg="Make sure there are no repeats in allowed_values. Values are not considered case-sensitive",
+            )
+        for val in allowed_values_set:
+            if val == "":
+                raise UserError("Empty strings are not allowed!")
```
pgm, line 199: Reading this error, I worry that it doesn't provide enough context. If I saw it when doing an upload, I might not realize that this was an error related to allowed_values.

How about the following?
```
raise UserError("Empty strings are not allowed in allowed_values!")
```

frontend/packages/@depmap/dataset-manager/src/components/DatasetForm.tsx
```diff
@@ -3,7 +3,7 @@ import * as React from "react";
 import { MatrixDatasetForm } from "./MatrixDatasetForm";
 import { TableDatasetForm } from "./TableDatasetForm";
 import { FormGroup, Radio } from "react-bootstrap";
-import { useState, useEffect, useMemo } from "react";
+import { useState, useEffect, useMemo, useCallback } from "react";
```
pgm, line 6: I'm guessing these are changes stemming from the branch being branched off of a different existing branch, correct?

pgm (approved):
Looks fine. Made some minor remarks and ignored all the ts/tsx changes because I'm assuming they're unrelated?

frontend/packages/@depmap/dataset-manager/src/components/DatasetForm.tsx
jessica-cheng: Yes I think I merged in a branch from [this PR](https://github.com/broadinstitute/depmap-portal/pull/142) to make testing on the UI side easier. I see it's approved now yay!

breadbox/breadbox/schemas/dataset.py
jessica-cheng: Good point!

PR #148, opened 2024-12-13
Fix topic forum link formatting and placement

jessica-cheng:
Fixed formatting where topic forum link after an image is misplaced and sizing is affected
The issue seems to be that the forum link was appended to the "last element" which could be a child element inside another outer element.


jessica-cheng:
@pgm  I will also merge this fix into skyros-24q4, dmc-24q4, and external-24q4 release branches. Please let me know if I'm missing anywhere else this fix should be merged into

pgm (approved):

pgm:
> @pgm I will also merge this fix into skyros-24q4, dmc-24q4, and external-24q4 release branches. Please let me know if I'm missing anywhere else this fix should be merged into

Just merging it into dmc-24q4 is fine. As part of the public release, I'll sync up all the branches.

PR #154, opened 2024-12-18
fix(breadbox): Addressing common portal error logs

snwessel:
Fixing two errors that I noticed were appearing frequently in the portal error logs:
1.  FastAPI is occasionally throwing a validation error when trying to construct a task response ([logs](https://console.cloud.google.com/errors/detail/CIPrkoS7xfmJ2QE;time=P30D;locations=global?project=cds-logging)): 
    ```
    {'type': 'int_from_float', 'loc': ('response', 'percentComplete'), 'msg': 'Input should be a valid integer, got a number with a fractional part', 'input': 0.03921781645880805}
    ```
2. The public portal was running into a lot of 500 errors when trying to load the compound_sensitivity tile on the cell line page. This was happening about 2,000 times per day! (Logs [here](https://console.cloud.google.com/errors/detail/CMW7v4G48dvgaw;time=P30D;locations=global?inv=1&invt=Abkd9A&project=cds-logging)). I believe this was happening because we don't include drug screen data in the public release. This endpoint will now return a 404 error instead. The frontend is already catching these errors and defaults to hiding the tile when an error is thrown. 

The 1st issue is the only one that might actually be creating problems for users, although custom analysis does seem to be working whenever I've tried it out (this issue might just be happening infrequently).

### DIFF
--- a/breadbox/breadbox/celery_task/utils.py
+++ b/breadbox/breadbox/celery_task/utils.py
@@ -180,7 +180,7 @@ def format_task_status(task):
         "id": task.id,
         "state": task.state,
         "message": message,
-        "percentComplete": int(percent_complete),
+        "percentComplete": int(percent_complete) if percent_complete else None,
         "nextPollDelay": 1000,  # units are miliseconds, this says once per second
         "result": result,
     }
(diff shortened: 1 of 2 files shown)


jessica-cheng (approved):
LGTM!

PR #160, opened 2025-01-03
feat(breadbox): Added storage of pre-computed associations to breadbox

pgm:
Adds endpoints for adding/deleting/querying pre-computed correlations between datasets to breadbox. (Which are then shown in data explorer as "precomputed associations")

This batch of changes only covers getting the required changes into breadbox. Once these are good to merge in, will proceed to add support to the breadbox loader and the necessary changes to the breadbox client.


breadbox/breadbox/service/slice.py
```diff
@@ -17,6 +20,63 @@
...
 
 
+@dataclass
+class ResolvedSlice:
+    dataset: Dataset
+    label: str
+    given_id: str
+
+
+def resolve_slice_to_components(
```
snwessel, line 30: This function sounds like it will be really helpful, but the naming made me a bit confused when I was first reading through. I tend to think of a "slice" being the full set of data belonging to the slice query, whereas I would call the label/given id "slice identifiers".  Would it be possible to rename this to something like `ResolvedSliceIdentifiers` or `DimensionIdentifiers` instead of `ResolvedSlice`?

breadbox/breadbox/service/metadata.py
```diff
@@ -162,6 +162,28 @@ def get_labels_for_slice_type(
         )
 
 
+def get_dataset_feature_label_by_id(
+    db: SessionWithUser, dataset: MatrixDataset, given_id: str
+):
+    """Looks up a single label for a given_id in a dataset. Internally this is implemented inefficiently so only
+    call this if you really want a single label. If you want to loop through many IDs, use get_matrix_dataset_sample_labels_by_id instead
+    (or we rewrite this function to make a DB query to fetch only the data we actually need)
+    """
```
snwessel, line 171: Excellent! I see what you mean now about this making it easier for us to improve performance later on.

breadbox/breadbox/api/temp/associations.py
```diff
@@ -0,0 +1,120 @@
...
+        serializer,
+        settings.compute_results_location,
+    )
+
+    dest_filename = f"associations/{uuid.uuid4()}.sqlite3"
+    full_dest_filename = os.path.join(settings.filestore_location, dest_filename)
+    # make sure the directory exists before trying to move the file to it's final name
+    os.makedirs(os.path.dirname(full_dest_filename), exist_ok=True)
+
+    os.rename(full_file, full_dest_filename)
```
snwessel, line 98: I see that in one of the tests, you explicitly define the schema for the little sqlite file/database, but you don't do that here (in the add_associations endpoint where the file is created). Is that because the user is expected to upload a sqlite3 file with the correct schema?

I suppose that would make sense given that we already have these files created somewhere else and just need to copy them into breadbox, right?

snwessel (approved):
Looks good to me! I just had a couple of minor questions/comments

breadbox/breadbox/service/slice.py
pgm: Sure, happy to rename this to `ResolvedSliceIdentifiers`

breadbox/breadbox/api/temp/associations.py
pgm: Yes, correct: my plan is to directly load the pre-existing sqlite3 files which are spit out from the preprocessing pipeline.

snwessel: Okay great!

PR #165, opened 2025-01-09
feat(breadbox): Improvements to breadbox client publishing

pgm:
Set of changes to make a manual job for publishing client and cleaning up publishing process


breadbox-client/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "breadbox-client"
-version = "3.20.0"
+version = "3.20.0+pgmdev3"
```
pgm, line 3: Ignore this. I'll change this back before merging in.

breadbox-client/pyproject.toml
```diff
@@ -10,7 +10,7 @@ packages = [
     {include = "breadbox_client"},
     {include = "breadbox_facade"},
 ]
-include = ["CHANGELOG.md", "breadbox_client/py.typed"]
+include = ["CHANGELOG.md", "breadbox_client/py.typed", "breadbox_client/**/*"]
```
pgm, line 13: to get poetry to override .gitignore's exclusions

.github/workflows/publish_breadbox_client.yaml
```diff
@@ -3,52 +3,24 @@ name: Publish breadbox client
...
-        with:
-          path: .venv
-          key: venv-${{ runner.os }}-${{ steps.full-python-version.outputs.version }}-${{ hashFiles('**/poetry.lock') }}
-      - name: Generate breadbox client
-        uses: ./.github/actions/generate-breadbox-client # defined as a re-usable action
-        # the above does a checkout which deletes the credential file, so we need to execute authenticate to google cloud after it
-      - name: Authenticate to Google Cloud
+
+      - name: "Authenticate to Google Cloud"
         uses: google-github-actions/auth@v2
```
snwessel, line 18: I'm guessing the warning below would go away if you updated this to `google-github-actions/auth@v2.1.7` (and did the same in `build_breadbox.yml`). 

And I'm curious if we added CodeQL to our repo, or is this something github added on their own? I've never seen a warning like that before

snwessel (approved):
Looks great, thank you for updating all of this! I think it'll be much nicer to have these smaller reusable actions. 

> However according to the poetry docs, it really sounds like ./bb update-package should NEVER be working and I can't explain why it is locally.

That's so interesting! At different moments in time, this command has or has not worked for me locally and I could never understand what changed in my local setup. I had thought there was something wrong with my local poetry setup that later got fixed 😅 I'm glad to hear this new command will be easier to use

.github/workflows/build_breadbox.yml
```diff
@@ -60,16 +60,23 @@ jobs:
...
+          fetch-depth: 0
+
+      - name: Prepare breadbox client
+        uses: ./.github/actions/prepare-breadbox-client # checks out and generates breadbox client code
+
+      - name: Check that the breadbox client passes pyright
         working-directory: ./breadbox-client
-        run: |
-          poetry install && poetry run pyright breadbox_facade
+        run: "run pyright breadbox_facade"
```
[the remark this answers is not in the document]
pgm: Yes, that's a mistake. Hmmm.. I'll have to look why that run skipped some steps...

.github/workflows/publish_breadbox_client.yaml
pgm: I didn't add it. My guess was maybe broad IT sec might have added it.

pgm: That's funny that adding a minor version would silence this warning... Adding a commit SHA is a cryptographic signature of the code ... but adding a minor version is a mutable tag like any other tag. I'm not sure how a minor version would prevent a supply chain attack. (That being said, I'm not especially worried about a supply chain attack from google, so I'm happy to do it just to silence the error)

Scutua (approved):

Scutua (approved):

Scutua (approved):

PR #168, opened 2025-01-14
feat(breadbox): Changed modeling of references between metadata

pgm:
Previously, the "references" field on column metadata was modeled in the database as an independent table which had FKs referencing datasets. These was marked as "delete on cascade" which caused the "references" field to be nulled out whenever the metadata dataset was deleted (as part of a dimension_type being updated).

These changes changes things so the relationship is stored on `tabular_column` instead, more closely mirroring the API, and the FK is to the dimension_type (which persists across updates) instead of dataset (which is deleted on updates to dimension types).

Also, changed `properties_to_index` to be stored on the `dimension_type` instead of having it have a FK to dataset, for similar reasons.


breadbox/breadbox/crud/types.py
```diff
@@ -503,9 +522,34 @@ def get_dimension_type_labels_by_id(
...
+    dimension_type = get_dimension_type(db=db, name=dimension_type_name)
+
+    if dimension_type is None:
+        raise ResourceNotFoundError(
+            f"Dimension type '{dimension_type_name}' not found. "
+        )
+    if dimension_type.dataset_id is None:
+        return {}
+
+    from ..models.dataset import Dimension
```
pgm, line 537: I forgot that I haven't yet removed these non-top-imports. There's some circular dependencies that need to be sorted out. 

Now that all the tests are passing, I'll go back an reorganize the files to avoid the circular imports. (I'm thinking I'll create a new "search_index.py" under service to collect all functions related to the search index.

snwessel: Excellent, thank you! I'd been thinking of making a `search_index` service a while back, but had decided to hold off until things were refactored. Hopefully that will help prevent a lot of future circular imports as well 🤞

breadbox/breadbox/models/dataset.py
```diff
@@ -290,68 +289,52 @@ class DimensionSearchIndex(Base, UUIDMixin, GroupMixin):
...
-        backref=backref(
-            "dimension_search_indexes",
-            cascade="all, delete-orphan",
-            passive_deletes=True,
-        ),
-    )
-    dimension_given_id = Column(String, nullable=False)
-    label = Column(String, nullable=False)
+    # The dimension which owns the property. (One should be able to group by dimension_id to get the bag of words associated with that dimension)
+    # dimension_id = Column(
```
snwessel, line 307: If this `dimension_id` column is being removed (which I agree makes sense), is there a reason to leave it commented out here?

breadbox/alembic/versions/e593fefbe9fc_move_column_refences.py
```diff
@@ -0,0 +1,126 @@
...
+        )
+
+    # run query to figure out which dimension records need updating
+    op.execute(
+        "create table tmp_references_to_copy as select dt_from.name datatype_name, dr.column column_name, from_col.id dimension_id, dt_to.name references_dimension_type_name from dimension_type dt_from join dataset dt_from_meta on dt_from_meta.id = dt_from.dataset_id join dataset_reference dr on dr.dataset_id = dt_from.dataset_id join dataset dt_to_meta on dt_to_meta.id = dr.referenced_dataset_id join dimension_type dt_to on dt_to.dataset_id = dt_to_meta.id join dimension from_col on (from_col.dataset_id = dt_from_meta.id and from_col.given_id = dr.column)"
+    )
+    # Do the actual updates
+    op.execute(
+        "update dimension set references_dimension_type_name = (select references_dimension_type_name from tmp_references_to_copy where id = dimension_id) where exists (select references_dimension_type_name from tmp_references_to_copy where id = dimension_id) "
+    )
```
snwessel, line 39: This migration looks good to me 👍

breadbox/breadbox/models/dataset.py
```diff
@@ -290,68 +289,52 @@ class DimensionSearchIndex(Base, UUIDMixin, GroupMixin):
...
+    #     ),
+    # )
     property = Column(String, nullable=False)
     value = Column(String, nullable=True)
-    priority = Column(Integer, nullable=True)
-    axis = Column(String, nullable=False)
-    type_name = Column(String, nullable=False)
-
-
-class DatasetReference(Base, UUIDMixin, GroupMixin, ReferencedGroupMixin):
```
snwessel, line 325: Fyi, now that this `DatasetReference` table is gone, you should be able to remove some of the complicated access control workarounds I'd added. Specifically in `breadbox/models/group.py` you can remove the definitions of:
* `ReferencedGroupMixin`
* `_add_filtering_criteria`

This is the only place those workarounds are used currently, and I can't think of anything else that would need them in the future. I believe we've decided that "dimension types" are always public, so there's no more need for the concept of "referenced groups".

breadbox/alembic/versions/e593fefbe9fc_move_column_refences.py
```diff
@@ -0,0 +1,126 @@
...
+    #
+    #
+    # with op.batch_alter_table('property_to_index', schema=None) as batch_op:
+    #     batch_op.add_column(sa.Column('dimension_type_name', sa.String(), nullable=False))
+    #     batch_op.drop_constraint('fk_property_to_index_dataset_id_dataset', type_='foreignkey')
+    #     batch_op.create_foreign_key(batch_op.f('fk_property_to_index_dimension_type_name_dimension_type'), 'dimension_type', ['dimension_type_name'], ['name'], ondelete='CASCADE')
+    #     batch_op.drop_column('dataset_id')
+    #
+
+    # drop search index and recreate it. We should add a step to deploy to populate this if it's empty
```
snwessel, line 85: I see your note here: 
>  drop search index and recreate it. We should add a step to deploy to populate this if it's empty

And I also see that you removed the command `populate-index` from `commands.py`. Is there another way to populate this search index during a deployment?

breadbox/breadbox/crud/types.py
```diff
@@ -503,9 +522,34 @@ def get_dimension_type_labels_by_id(
     return get_dimension_type_metadata_col(db, dimension_type_name, col_name="label")
 
 
+def get_given_id_to_dimension_id_mapping(
```
snwessel, line 525: Does this function actually get used somewhere? 

We'd briefly talked about how you realistically only need the `given_id` to perform most joins (not the `dimension_id`). When I saw this defined here, I was curious to see where we still need a mapping like this, but I can't find anywhere that this function is actually used (only one place that it's imported) 🤔 I may just be missing it though.

breadbox/breadbox/crud/dataset_reference.py
```diff
@@ -2,54 +2,6 @@
 
 from breadbox.db.session import SessionWithUser
 from ..models.dataset import (
-    DatasetReference,
     DimensionType,
     TabularDataset,
 )
```
snwessel, line 7: Could this file just be deleted now? It looks like the imports are the only part that haven't already been removed.

snwessel (approved):
Overall, looks great to me! Just left a few questions/comments on specific things I was confused about

breadbox/breadbox/models/dataset.py
pgm: Nope -- I'll clean it up.

breadbox/alembic/versions/e593fefbe9fc_move_column_refences.py
pgm: Ah, right. I forgot to circle back and do that part. 

My plan was to add a step that's always run as part of migrations. The new step would check to see if the index is empty, and if so, re-run it on all dimension_types. If the table _isn't_ empty, then it'd do nothing. 

(It would allow us to "invalidate" the index by deleting the contents. I like this for the simplicity. I don't like this for how implicit it is... For example, I don't think it'll be obvious to anyone that this behavior exists. Anyway, I'll give it a try because I don't have better ideas.)

breadbox/breadbox/crud/types.py
pgm: Oh, that's left over from some earlier (failed) changes. I thought I had removed it. 

...I think it gets removed in my other re-org branch for when I moved things into services. Anyway, I'll confirm and yank it if it's not already removed.

breadbox/breadbox/crud/dataset_reference.py
pgm: Yes, good point. I'll delete.

breadbox/breadbox/crud/types.py
pgm: yes, confirmed this method is gone in the other branch.

PR #170, opened 2025-01-15
Initial draft of the Data Prep Pipeline

naquib314:
This is the initial draft of the data prep pipeline. It can be run both locally(instructions in [README.md](https://github.com/broadinstitute/depmap-portal/blob/data-prep-pipeline/data-prep-pipeline/README.md)) or in Jenkins(https://datascidev.broadinstitute.org/job/Data%20Prep%20Pipeline%201.0/). At this moment it produces the following files and uploads them to Taiga:

**Release Data:**
- PortalOmicsCNGeneLog2
- Gene
- PortalCompounds

**Predictability Data**
- GeneticDerangement
- DriverEvents
- Fusion
- Lineage
- CRISPRConfounders

**Legacy Data**
- RNAiConfounders
- RNAiDep
- OncRefConfounders
- RepSinglePtConfounders
- Metabolomics


data-prep-pipeline/data_prep_pipeline/common.conseq
```diff
@@ -0,0 +1,17 @@
...
+
+include "data_prep_pipeline/release_inputs-DO-NOT-EDIT-ME"
+
+include "data_prep_pipeline/update_hgnc_gene_table.conseq"
+include "data_prep_pipeline/cngene_log_2_transformation.conseq"
+include "data_prep_pipeline/filter_portal_compounds.conseq"
+include "data_prep_pipeline/predictability.conseq"
+include "data_prep_pipeline/legacy_data.conseq"
+
+include "data_prep_pipeline/publish.conseq"
```
pgm, line 17: This looks like this always runs. I thought we had talked about only running the publish rules if it was run from jenkins, because as-is, if a developer runs this file for testing purposes, won't it overwrite files in the release?

In`_run_common.conseq` in the preprocessing pipeline, we have the following to achieve this:

```
# all of the run_xxx.conseq files define publish_dest as INVALID because
# this variable is overriden when run from jenkins. If the valid is still
# INVALID, then we're probably running the pipeline for development, and
# therefore, don't publish any of the results to the dest path
if "config.get('publish_dest', 'INVALID') != 'INVALID'":
  include "publish.conseq"
endif
```
...but I don't see anything like that here.

data-prep-pipeline/data_prep_pipeline/filter_portal_compounds.conseq
```diff
@@ -0,0 +1,14 @@
+rule filter-portal-compounds:
+    inputs:
+        repsdrug_matrix={"type": "repsdrug_matrix"},
```
pgm, line 3: We should also include GDSC1, GDSC2, and CTD^2 datasets to use for the filtering. I have a hunch the taiga IDs for these are hardcoded in filter_portal_compounds.py, but it'd be good to lift them out so they're in the conseq script.

From early experiences with the preprocessing pipeline, we learned it really is too easy to end up using an old/stale taiga ID when we put the IDs in the scripts themselves. As a result, we ended up with a policy of putting all taiga IDs into a single file.

data-prep-pipeline/data_prep_pipeline/update_hgnc_gene_table.conseq
```diff
@@ -0,0 +1,18 @@
...
+            "type": "gene",
+            "filename": {"$filename": "hgnc_gene_table.csv"},
+            "orig_dataset_id": "{{inputs.hgnc_gene_table.dataset_id}}"
+        }
+    run "python3" with """
+        from taigapy import create_taiga_client_v3
+
+        tc = create_taiga_client_v3()
+        print("Getting HGNC gene data...")
+        hgnc_gene_df = tc.get("{{inputs.hgnc_gene_table.dataset_id}}")
```
pgm, line 15: Instead of downloading the CSV and uploading a new CSV, can you add the file to destination with a `TaigaReference`? That way taiga will track the provenance of the original file, and we won't end up with two taiga datasets which are identical.

data-prep-pipeline/gsutil-boto-config
```diff
@@ -0,0 +1,370 @@
+# This file contains credentials and other configuration information needed
```
pgm, line 1: What is this file for?

data-prep-pipeline/README.md
```diff
@@ -5,37 +5,30 @@ The data prep pipeline gets data from **Taiga** and prepares the data to make th
...
-3. Import and add that script to the main function of [data prep.py](data_prep_pipeline/data_prep.py).
+Then, assuming you are in `depmap-portal/data-prep-pipeline` where this readme is located, install and activate a poetry environment. Then:
 
-An example can be found [here](data_prep_pipeline/predictability/transform_crispr_confounders.py).
+1. Run `poetry shell`.
+2. Once inside the poetry environment, run `data_prep_pipeline/common.conseq` which will run each rule mentioned there and produce the relevant output.
 
-#### If source data is from release+some other dataset in taiga
+Note that there are two primary configuration files, `release_inputs.conseq` which contains all the taiga ids of the initial inputs for different rules and `common.conseq` which contains all the available rules. There's a `data_prep_pipeline/publish.conseq` file where each upload to taiga is configured and executed. If you would like to modify or skip the upload for a particular rule, then do so in that file.
```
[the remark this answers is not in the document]
naquib314: Sounds good.

data-prep-pipeline/README.md
```diff
@@ -5,37 +5,30 @@ The data prep pipeline gets data from **Taiga** and prepares the data to make th
...
 
-In case of the other dataset in taiga, add that to the [config.py](data_prep_pipeline/config.py) as a variable and import that in your transformation script. Everything else is exactly as [If Source Data is from release only](#if-source-data-is-from-release-only).
+## Run the Data Prep Pipeline in Jenkins
 
-An example can be found [here](data_prep_pipeline/predictability/transform_fusion.py)
+Go to the Data Prep Pipeline 1.0 jenkins job here: https://datascidev.broadinstitute.org/job/Data%20Prep%20Pipeline%201.0/
```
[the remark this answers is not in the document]
naquib314: I am not sure why. Perhaps we can chat about it during our check-in?

data-prep-pipeline/README.md
```diff
@@ -5,37 +5,30 @@ The data prep pipeline gets data from **Taiga** and prepares the data to make th
...
 
-#### If source data is from legacy only
+Then click on build. Optionally choose one of the parameters if you would like a clean start or start with a specific export or want to automatically rebuild the db once done.
 
-We are not expecting many such datasets to be out there. But if you need to add one, then:
+## How to add a new dataset
```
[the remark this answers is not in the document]
naquib314: Right, I will change the wording.

data-prep-pipeline/data_prep_pipeline/common.conseq
naquib314: We did. This was a mistake on my part. 
I will write that logic so that a local run doesn't  overwrite data in release.

data-prep-pipeline/gsutil-boto-config
naquib314: This should not be here. I will delete it.

data-prep-pipeline/data_prep_pipeline/filter_portal_compounds.conseq
naquib314: Need to chat about this in check in.

data-prep-pipeline/data_prep_pipeline/legacy_data.conseq
```diff
@@ -0,0 +1,96 @@
...
+        tc = create_taiga_client_v3()
+        print("Getting RNAi confounders data...")
+        rnai_confounders_matrix = tc.get("{{inputs.rnai_confounders.dataset_id}}")
+        rnai_confounders_matrix.set_index("Row.name", inplace=True)
+        rnai_confounders_matrix.index.name = None
+        rnai_confounders_matrix.to_csv("rnai_confounders.csv")
+    """
+
+
+rule update_rnai_data:
```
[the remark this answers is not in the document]
naquib314: Chatted and have clear guidelines now.

PR #171, opened 2025-01-16
Replace Vector Catalog in Cell Line Selector (cherry-picked)

rcreasi:
This commit already exists on `master`. This just cherry-picks it onto the `external-24q4` branch.

### DIFF
--- a/frontend/packages/@depmap/data-explorer-2/src/utils/misc.ts
+++ b/frontend/packages/@depmap/data-explorer-2/src/utils/misc.ts
@@ -31,6 +31,10 @@ export const isCompleteExpression = (expr: any) => {
     return false;
   }
 
+  if (typeof expr === "boolean") {
+    return true;
+  }
+
   if (expr.and && expr.and.length === 0) {
     return false;
   }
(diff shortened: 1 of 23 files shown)


frontend/packages/@depmap/data-explorer-2/src/utils/misc.ts
```diff
@@ -91,6 +95,34 @@ export const isSampleType = (dimensionType: string | null | undefined) => {
...
+
+  const expr = dimension.context.expr as { "==": [object, string] };
+  const feature = expr["=="][1];
+
+  return [
+    "slice",
+    urlLibEncode(dimension.dataset_id),
+    urlLibEncode(feature),
+    "label",
+  ].join("/");
```
rcreasi, line 123: @snwessel This is where I construct a legacy slice ID. The public branch is pretty far behind master so this will probably be prone to the same bug you just fixed with Breadbox datasets. I think I can handle that here instead. What format should I use for that? `"breadbox/<dataset-id>/<feature>/label"` ?

rcreasi: Oh wait, that's only relevant to custom analysis, right? This is just in Cell Line Selector that doesn't use any of those endpoints.

snwessel: Yeah, that bug was specific to custom analysis which uses some older endpoints that don't fully use the `data_access` interface. I'm not super familiar with how custom analysis works though, so it might be good to test these changes with a breadbox dataset before deploying.

snwessel (approved):
Looks good to me 👍

PR #174, opened 2025-01-23
Predictability Initial Draft

naquib314:
This is the very first draft of predictability as part of conseq in the new analysis pipeline.
There are mainly two files. One is `model-config.yaml`, which is used as the blueprint of model inputs and the other is the `fit.conseq`.
`fit.conseq` works as follows:
1. It first creates model input json files based on the `model-config.yaml` file.
2. Once the input json file is created, daintree is run to produce the output and upload them to taiga. There's an additional output config file that is created. 
3. Combine the output config files into one file. 

TODO:

- [x] Once `fit.conseq` is producing the right output config file which includes the input json file's content and the destination taiga IDs of the produced outputs(e.g. ensemble.csv, metadata, etc.), the `model-config.yaml` file and the `fit.conseq` needs to be updated to reflect all 5 models for CRISPR and RNAi.

- [x] This implementation works. However, I wonder if **I could produce multiple outputs(input json files) in one rule and then use them to run the models parallelly in another rule**.


analysis-pipeline/predictability/fit.conseq
```diff
@@ -0,0 +1,134 @@
...
+        output_file_path = os.path.abspath(output_file)
+
+        with open(output_file, 'w') as f:
+            json.dump(output_json, f, indent=2)
+        print(f"Created {output_file}")
+
+        
+        docker_command = [
+            "docker", "run",
+            "-w", "/daintree",
```
pgm, line 104: If you set the working dir in the dockerfile you won't need this.

analysis-pipeline/predictability/fit.conseq
```diff
@@ -0,0 +1,134 @@
...
+            "--pull=always",
+            "-v", f"{output_file_path}:/daintree/{output_file}",
+            "us.gcr.io/broad-achilles/daintree-sparkles:v4",
+            "/install/depmap-py/bin/python3.9", "-u", "run_fit_models.py",
```
pgm, line 108: What do you think about putting in the dockerimage an executable script in the path.

Feels like it'd be simpler if we were just running something like `daintree collect-and-fit-generated-config ...`

analysis-pipeline/predictability/fit.conseq
```diff
@@ -0,0 +1,134 @@
...
+            "collect-and-fit-generate-config",
+            "--input-files", output_file,
+            "--sparkles-path", "/install/sparkles/bin/sparkles",
+            "--sparkles-config", "/daintree/sparkles-config",
+            "--save-dir", "/daintree/output_data/nayeem-test",
```
pgm, line 113: I don't see /daintree/output_data being mounted from outside of the docker image. That suggests the output will get thrown away when the container is deleted.

Instead of mounting individual files, I often find it easier to mount the current working directory to some location in the container and use that for input and output.

analysis-pipeline/predictability/fit.conseq
```diff
@@ -0,0 +1,134 @@
...
+            "docker", "run",
+            "-w", "/daintree",
+            "--pull=always",
+            "-v", f"{output_file_path}:/daintree/{output_file}",
+            "us.gcr.io/broad-achilles/daintree-sparkles:v4",
+            "/install/depmap-py/bin/python3.9", "-u", "run_fit_models.py",
+            "collect-and-fit-generate-config",
+            "--input-files", output_file,
+            "--sparkles-path", "/install/sparkles/bin/sparkles",
+            "--sparkles-config", "/daintree/sparkles-config",
```
pgm, line 112: Where is this coming from? Is this in the docker image? I don't think it should be because it's configuration which will vary by who is using this image.

analysis-pipeline/predictability/fit.conseq
```diff
@@ -0,0 +1,134 @@
...
+            )
+            
+            print("STDOUT:", result.stdout)
+            print("STDERR:", result.stderr)
+
+        except subprocess.CalledProcessError as e:
+            print(f"Command failed with return code {e.returncode}")
+            print("STDOUT:", e.stdout)
+            print("STDERR:", e.stderr)
+    """
```
pgm, line 134: Concretely, if you were to take all my suggestions this would simplify down to:
```
        docker_command = [
            "docker", "run",
            "-v", f"{os.getcwd()}:/work",
            "us.gcr.io/broad-achilles/daintree-sparkles:v4",
            "daintree",
            "collect-and-fit-generate-config",
            "--input-files", f"/work/{output_file}",
            "--sparkles-config", "/work/sparkles-config",
            "--save-dir", "/work/output_data/nayeem-test",
            "--test", "True",
        ]

        subprocess.run(
            docker_command,
            check=True,
        )

File: analysis-pipeline/predictability/fit.conseq [hunk 6]
```diff
@@ -0,0 +1,134 @@
...
+        docker_command = [
+            "docker", "run",
+            "-w", "/daintree",
+            "--pull=always",
+            "-v", f"{output_file_path}:/daintree/{output_file}",
+            "us.gcr.io/broad-achilles/daintree-sparkles:v4",
+            "/install/depmap-py/bin/python3.9", "-u", "run_fit_models.py",
+            "collect-and-fit-generate-config",
+            "--input-files", output_file,
+            "--sparkles-path", "/install/sparkles/bin/sparkles",
```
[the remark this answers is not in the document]
naquib314: It is. I will take it out.

analysis-pipeline/predictability/fit.conseq
```diff
@@ -0,0 +1,134 @@
...
+
+        with open(output_file, 'w') as f:
+            json.dump(output_json, f, indent=2)
+        print(f"Created {output_file}")
+
+        
+        docker_command = [
+            "docker", "run",
+            "-w", "/daintree",
+            "--pull=always",
```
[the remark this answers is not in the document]
naquib314: You are right. It was there due to part of the development. I will take it out.

analysis-pipeline/predictability/fit.conseq
```diff
@@ -0,0 +1,134 @@
...
+        confounder={"type": "feature", "label": "confounder"},
+        driver_events={"type": "feature", "label": "driver_events"},
+        model_config=fileref("model-config.yaml")
+    outputs:
+        {
+        "type": "input-model-file",
+        "filename": {"$filename": "model-map.json"}
+        },
+
+    run "python" with """
```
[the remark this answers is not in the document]
naquib314: Will need to chat about some strategy about this during our meeting.

analysis-pipeline/predictability/fit.conseq
```diff
@@ -0,0 +1,134 @@
...
+            "--pull=always",
+            "-v", f"{output_file_path}:/daintree/{output_file}",
+            "us.gcr.io/broad-achilles/daintree-sparkles:v4",
+            "/install/depmap-py/bin/python3.9", "-u", "run_fit_models.py",
+            "collect-and-fit-generate-config",
+            "--input-files", output_file,
+            "--sparkles-path", "/install/sparkles/bin/sparkles",
+            "--sparkles-config", "/daintree/sparkles-config",
+            "--save-dir", "/daintree/output_data/nayeem-test",
+            "--test", "True",
```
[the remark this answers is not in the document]
naquib314: It defaults to False now. I will update the daintree doc.

analysis-pipeline/predictability/fit.conseq
```diff
@@ -0,0 +1,134 @@
...
+    
+    # Process each model in the YAML
+    for model_name, model_config in config.items():
+        output_json = {
+            "model_name": model_name,
+            "screen_name": model_config["Screen"],
+            "data": {}
+        }
+        
+        target = model_config["Target"][0]
```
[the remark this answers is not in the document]
naquib314: This is getting the first element of each Target. But I changed the yaml file as I added more dataset and will be talking to you about some strategy.

analysis-pipeline/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+You just need conseq installed since everything else that is required is installed inside the us.gcr.io/broad-achilles/daintree-sparkles:v4 image where predicability is run.
+
+Note that there is a model-config.yaml file which has the config of all the models.
+Once conseq is installed, you can run `conseq run fit.conseq` to start.
+
+The `fit.conseq` works as follows:
+
+1. It first creates model input json files based on the `model-config.yaml` file.
+2. Once the input json file is created, daintree is run to produce the output for predictability.
```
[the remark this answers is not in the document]
naquib314: Yes, that is correct.

analysis-pipeline/predictability/fit.conseq
naquib314: Yes, I thought about it before and it makes sense. Will do that.

analysis-pipeline/predictability/fit.conseq
naquib314: Sounds good.

analysis-pipeline/predictability/fit.conseq
```diff
@@ -0,0 +1,155 @@
...
+
+        # Script to generate daintree input config file
+        script=fileref("scripts/generate_daintree_input_configs.py"),
+
+    run "python" with """
+    import json
+
+    config_dict = {}
+
+    for key, value in {{inputs}}.items():
```
pgm, line 39: I'm not really sure why this loop is needed to construct the dictionary.

Can't we just assign the dict to `config_dict`? 

That is to say:
```
config_dict = {{inputs}}
```

analysis-pipeline/predictability/fit.conseq
```diff
@@ -0,0 +1,155 @@
...
+    run "python" with """
+      import subprocess
+      import os
+      import glob
+      import json
+
+      input_config_filepath = "{{ inputs.daintree_input_config.filename }}"
+      input_config_filename = "{{ inputs.daintree_input_config.label }}.json"
+
+      docker_command = [
```
pgm, line 71: ah, much easier to read

analysis-pipeline/predictability/fit.conseq
```diff
@@ -0,0 +1,155 @@
...
+        "--upload-to-taiga", "predictability-76d5",
+      ]
+
+      subprocess.run(
+          docker_command,
+          check=True
+      )
+      
+      # Find the output config file using glob
+      output_config_files = glob.glob(os.path.join(os.getcwd(), "output_data", "output_config_files", "*.json"))
```
pgm, line 93: It's surprising that we need to search by wildcard. Does daintree write out files with non-predictable filenames?

pgm (approved):

analysis-pipeline/predictability/fit.conseq
naquib314: At this moment, it produces a name like `OutputConfigCellContextCRISPR{model_name}{screen_name}.json`. I can of course change that to something like `output_config.json` and use that here. Will do that.

PR #175, opened 2025-01-23
Compound dataset refactoring: cell line page 

snwessel:
**Goal**: Update the cell line page (specifically the Preferential dependency tiles) so that it will work automatically with either:
* PRISM data stored in breadbox (indexed by compound)
* PRISM data stored in the legacy backend (indexed by compound experiment)

I decided to start the compound-related refactoring with this part of the codebase because: 
1. I'm pretty familiar with it
2. It needed to be refactored at some point anyway since it was tightly coupled with the legacy backend
3. This feature does already display compound-level data to users (so the UI doesn't need to change at all)

**The main changes I made here were:**
1. Defining a new method on the `data_access` interface: `get_subsetted_df_by_compound_labels` 
    * This is very similar to our existing function `get_subsetted_df_by_labels` except that when called on a legacy dataset indexed by compound experiment, it will re-index it by compound. 
2. I did some refactoring of the cell line page endpoints to call this new method
3. I moved the compound-experiment-specific functionality to `depmap/compound/legacy_utils.py` where it can be clearly deprecated and hidden behind this new interface 

After refactoring, I confirmed that the existing tests are still passing and that the UI is appearing identically to before (at least with dev data) 
image

**Any feedback is appreciated!** As I work on other parts of the codebase, I will continue to follow this pattern wherever it makes sense.

### DIFF
--- a/portal-backend/depmap/compound/legacy_utils.py
+++ b/portal-backend/depmap/compound/legacy_utils.py
@@ -31,7 +31,7 @@ def get_compound_labels_for_compound_experiment_dataset(dataset_name: str) -> di
     )
     return {experiment_id: compound_label for experiment_id, compound_label in labels_by_indeces}
 
-def get_dataset_data_indexed_by_compound_label(dataset_id: str) -> pd.DataFrame:
+def get_subsetted_df_by_compound_labels(dataset_id: str) -> pd.DataFrame:
     """
     Load the data for a drug screen dataset. This is similar to get_subsetted_df_by_labels,
     except that for compound datasets, the result will be indexed by compound (to match breadbox).
--- a/portal-backend/depmap/compound/legacy_utils.py
+++ b/portal-backend/depmap/compound/legacy_utils.py
@@ -34,7 +34,7 @@ def get_compound_labels_for_compound_experiment_dataset(dataset_name: str) -> di
 
 def get_dataset_data_indexed_by_compound_label(dataset_id: str) -> pd.DataFrame:
     feature_type = interactive_utils.get_entity_type(dataset_id)
-    assert feature_type == "compound", f"Cannot re-index a non-compound dataset '{dataset_id}' by compound label"
+    assert feature_type == "compound_experiment", f"Dataset '{dataset_id}' is indexed by '{feature_type}', cannot be re-indexed by compound label"
     compound_labels_by_experiment = get_compound_labels_for_compound_experiment_dataset(dataset_id)
     compound_experiment_df = interactive_utils.get_subsetted_df_by_labels(dataset_id, None, None)
     compound_df = compound_experiment_df.rename(index=compound_labels_by_experiment)
--- a/portal-backend/depmap/compound/legacy_utils.py
+++ b/portal-backend/depmap/compound/legacy_utils.py
@@ -1,0 +1,27 @@
+import sqlalchemy as sa
+
+from depmap.interactive import interactive_utils
+from depmap.dataset.models import (
+    Compound,
+    CompoundExperiment,
+)
+from depmap.partials.matrix.models import Matrix, RowMatrixIndex
+
+def get_compound_labels_for_compound_experiment_dataset(dataset_name: str) -> dict[str, str]:
+    """
+    In the future, all drug screen datasets will be indexed by compound instead of compound experiment. 
+    This method loads a mapping between the old index (compound experiment label) and 
+    method is used to re-index datasets which are indexed by compound experiment. 
+    """
+    matrix_id = interactive_utils.get_matrix_id(dataset_name)
+    comp_exp_alias = sa.orm.aliased(CompoundExperiment)
+    compound_alias = sa.orm.aliased(Compound)
+    labels_by_indeces = (
+        Matrix.query.filter_by(matrix_id=matrix_id)
+        .join(RowMatrixIndex)
+        .join(comp_exp_alias)
+        .join(compound_alias, compound_alias.entity_id == comp_exp_alias.compound_id)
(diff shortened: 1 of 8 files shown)


portal-backend/depmap/data_access/interface.py
```diff
@@ -399,36 +410,13 @@ def get_custom_cell_lines_dataset() -> str:
...
-
-
-def is_standard(dataset_id: str) -> bool:
-    """
-    Check whether the given dataset is standard or nonstandard. 
-    Only applicable for the legacy data access implementation.
-    """
-    return interactive_utils.is_standard(dataset_id)
-
-
```
snwessel, line 431: Unrelated: a bunch of these deprecated functions aren't being used anymore 🎉 (at least in the modules that reference the `data_access` interface) so I'm removing them from the interface.

portal-backend/depmap/compound/legacy_utils.py
```diff
@@ -0,0 +1,49 @@
...
+def get_subsetted_df_by_compound_labels(dataset_id: str) -> pd.DataFrame:
+    """
+    Load the data for a drug screen dataset. This is similar to get_subsetted_df_by_labels,
+    except that for compound datasets, the result will be indexed by compound (to match breadbox).
+    """
+    feature_type = interactive_utils.get_entity_type(dataset_id)
+    assert feature_type == "compound_experiment", f"Dataset '{dataset_id}' is indexed by '{feature_type}', cannot be re-indexed by compound label"
+    compound_labels_by_experiment = get_compound_labels_for_compound_experiment_dataset(dataset_id)
+    compound_experiment_df = interactive_utils.get_subsetted_df_by_labels(dataset_id, None, None)
+    compound_df = compound_experiment_df.rename(index=compound_labels_by_experiment)
```
pgm, line 43: If you've verified that this all works, then sounds fine with me.

But just in case it's useful to hear an alternative approach: The strategy I tend to use for something like this tends to be a little different. As opposed to de-duping at the end, I tend to create an index based on what I want and then subset by that.

Concretely, it'd look something like:
1. instead of `get_compound_labels_for_compound_experiment_dataset` returning all compound experiment IDs, I'd have the collapse happen there. (ie: the dictionary would be something like `max(experiment_id) -> compound_label` (where the `max` is per compound_label, so that compound_label appears once in the dict)
2. then subset using the keys of `compound_labels_by_experiment`

step #2 turns:

```
    compound_df = compound_experiment_df.rename(index=compound_labels_by_experiment)
    # Check for duplicate compound labels (which is possible since there are multiple CEs per compound)
    if compound_df.index.duplicated().any():
        log.warning(f"Found duplicate compounds in the dataset {dataset_id}. Keeping first occurance, dropping others.")
        compound_df = compound_df.reset_index().drop_duplicates(subset="index").set_index("index") # pyright: ignore
```

into:
```
compound_df = compound_experiment_df.loc[list(compound_labels_by_experiment.keys()),:].rename(index=compound_labels_by_experiment)
```

I think my gut feeling comes from in the original implementation, there's a brief window of time in which the index of the data frame has duplicated entires and that violates one of the assumptions I typically make about data frames. My version is trying to avoid that but never constructing a dataframe with dups in the index.

It's a short window, it's really fine, so I think it's just a matter of personal taste.

portal-backend/depmap/data_access/interface.py
```diff
@@ -308,6 +299,26 @@ def get_slice_data(slice_query: SliceQuery) -> pd.Series:
 
     else:
         raise Exception("Unrecognized slice query identifier type")
+    
+###############################################################
+# METHODS BELOW ARE SPECIAL WORKAROUNDS FOR COMPOUND DATASETS #
+###############################################################
+# In the future, all drug screen datasets will be indexed by compound instead of compound experiment.
+# These methods exist to ensure that both the legacy backend and breadbox are returning the same
```
pgm, line 307: I'd say `same shaped data`. There's some arbitrariness in the choice of which compound experiment gets picked in this transitionary period.  That being said, I appreciate the comments here capturing the motivation!

portal-backend/pyright-ratchet-errors.txt
```diff
@@ -668,8 +667,6 @@ views.py: error: Argument of type "Literal['value']" cannot be assigned to param
 views.py: error: Argument of type "Literal[DependencyEnum.Avana]" cannot be assigned to parameter "dependency_dataset_name" of type "str" in function "get_dataset_by_name"
 views.py: error: Argument of type "Unknown | FileStorage" cannot be assigned to parameter "filepath_or_buffer" of type "FilePath | ReadCsvBuffer[bytes] | ReadCsvBuffer[str]" in function "read_csv" (reportArgumentType)
 views.py: error: Argument of type "Unknown | Hashable" cannot be assigned to parameter "group_name" of type "str" in function "__init__"
-views.py: error: Argument of type "Unknown | None" cannot be assigned to parameter "cell_line_col_index" of type "int" in function "get_all_cell_line_compound_sensitivity"
```
pgm, line 671: fewer typing errors!

pgm (approved):
Looks great to me. I like the strategy and wonder if this could be used, with some small tweaks, to cover many other cases.

portal-backend/depmap/compound/legacy_utils.py
snwessel: Thank you! That does sound like a better implementation - and it helps more generally with some of the awkwardness of mapping between compounds and compound experiments. 

As I started looking into refactoring for the compound page, I realized that some of these "legacy utils" will need to be a bit more generalizable, so I'll probably need to update them anyway. 

I am going to merge this in as is for now, but I will incorporate this feedback in my next changes 👍

PR #176, opened 2025-01-24
fix(breadbox): Improve tabular dataset error handling

jessica-cheng:
### Issue 1
The issue fixed was that the celery worker ran into a `ValueError` when trying to upload a tabular dataset instead of an `HTTPException` which the celery task result handler couldn't process into a meaningful error message to the client.

Screenshot 2025-01-24 at 12 09 19 PM

### Issue 2
While fixing the above issue, another issue discovered was that although we had a validation check for the dimension type id column missing in the dataset file upload, it was not being caught since the validation happens AFTER we try to access the missing id column in the file in the code. I did some refactoring since there were multiple places we handle validation for tabular datasets and cleaned up the logic to ensure validation to dataset occurs earlier.

Screenshot 2025-01-24 at 2 18 40 PM
Screenshot 2025-01-24 at 2 45 37 PM


pgm (approved):
Looks good to me.

PR #180, opened 2025-01-27
remove Elara version of GroupsPage

rcreasi:
We have a shared module (@depmap/groups-manager) that is identical. No sense it having two copies ¯\\\_(ツ)_/¯


frontend/packages/portal-frontend/src/groupsManager/components/GroupsManager.tsx
```diff
@@ -4,7 +4,7 @@ import { ApiContext } from "@depmap/api";
 import { useCallback, useContext, useEffect, useState } from "react";
 import { GroupArgs, GroupEntryArgs } from "@depmap/types";
 
-import { GroupsPage } from "@depmap/groups-manager";
```
rcreasi, line 7: https://react.dev/reference/react/lazy#:~:text=exported%20as%20the%20default%20export

frontend/packages/@depmap/groups-manager/index.ts
```diff
@@ -1,3 +1,3 @@
-export { default as GroupsPage } from "./src/components/GroupsPage";
+export { default } from "./src/components/GroupsPage";
```
rcreasi, line 1: Now exporting `GroupsPage` as the default export so it can be used with `React.lazy()` [here](https://github.com/broadinstitute/depmap-portal/pull/180/files#diff-4930847b0087c2d260e5e06e73c90fbbf05d2afc337105a70b45fdb936e431dbR24).

https://react.dev/reference/react/lazy#:~:text=exported%20as%20the%20default%20export

jessica-cheng: Thanks!

PR #181, opened 2025-01-27
Fix error detail accessor

jessica-cheng:
When debugging an error, I encountered a different error when trying to upload a dataset with bad request params. The error returned provided a vague error message. Looking into this more, it looks like the upload dataset function encountered an error before the task was passed to celery resulting in a non-celery error being uncaught on the frontend. We already had a function that checks if a response object is an error but its property accessors were incorrect presumably due to earlier deprecated changes.

### Fix catches non-celery error and correctly shows error message in UI
Screenshot 2025-01-23 at 6 11 02 PM


pgm (approved):
I'm not a good judge for this one, but I don't see anything concerning and I trust you tried this out. :)

PR #183, opened 2025-01-29
fix(breadbox): Refactor tabular dataset schema error

jessica-cheng:
Error message returned to client is an obscure pandera error. Change error message to be more intuitive to users

This PR addresses the original task's use case. While completing this task, I noticed many pandera error codes are obscure and that stringifying the SchemaError often gives a better error message. However, in some cases doing so messes up the formatting of the message in the frontend so that is a TODO that I will address later. In addition, it is difficult to catch every single schema error case but I have noted down the types of errors I have accounted for and expect the schema validation to catch so far..

Screenshot 2025-01-29 at 4 13 07 PM

### DIFF
--- a/breadbox/breadbox/io/data_validation.py
+++ b/breadbox/breadbox/io/data_validation.py
@@ -628,7 +628,7 @@ def _validate_tabular_df_schema(
     )
 
     # NOTE: missing values are denoted as pd.NA
-    df = pd.read_csv(file_path)
+    df = pd.read_csv(file_path, dtype=str)
     try:
         validated_df = schema.validate(df)
     except SchemaError as schema_error:
--- a/breadbox/breadbox/io/data_validation.py
+++ b/breadbox/breadbox/io/data_validation.py
@@ -8,7 +8,7 @@ import numpy as np
 import pandas as pd
 from pandas.api.types import is_numeric_dtype, is_string_dtype, is_bool_dtype
 import pandera as pa
-from pandera.errors import SchemaError
+from pandera.errors import SchemaError, SchemaErrorReason
 from fastapi import UploadFile
 from sqlalchemy import and_, or_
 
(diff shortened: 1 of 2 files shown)


breadbox/breadbox/io/data_validation.py
```diff
@@ -613,43 +613,30 @@ def get_checks_for_col(annotation_type: AnnotationType):
     schema = pa.DataFrameSchema(
         {
             k: pa.Column(
-                annotation_type_to_pandas_column_type(
-                    v.col_type
-                ),  # annotation_type_to_pandera_column_type(v.col_type),
-                nullable=False if dimension_type_identifier == k else True,
+                annotation_type_to_pandas_column_type(v.col_type),
+                coerce=True,  # SchemaErrorReason: DATATYPE_COERCION
```
jessica-cheng, line 617: Instead of reading csv into df with expected dtype, check within the schema so that values in df column that are not of the assigned dtype raises a `SchemaError`

pgm (approved):
LGTM

breadbox/breadbox/io/data_validation.py
pgm: Thinking about this more, I do still have a bit of worry about this. 

I think there's an edge case. I'd be most comfortable if you read all columns saying to expect "str" and then used the validator to coerce them into the right type.

I worry about unexpected conversions and the way that there might be some mangling along the way.

Here's a concrete example, image a csv which looks like:
```
v
1
2.1
```

Now, for column "v", I'd expect the values to be `["1", "2.1"]` but this process will yield `["1.0", "2.1"]`.

Here's the code I used to verify this behavior:
```
>>> df = pd.read_csv(io.StringIO("v\n1\n2.1"))
>>> list(pa.DataFrameSchema({"v": pa.Column(str)}, coerce=True).validate(df)["v"])
['1.0', '2.1']
```

If you read in all columns as strings first, then there's no lossy transform and I think the problem is avoided:
```
>>> df = pd.read_csv(io.StringIO("v\n1\n2.1"), dtype=str)
>>> list(pa.DataFrameSchema({"v": pa.Column(str)}, coerce=True).validate(df)["v"])
['1', '2.1']
```

jessica-cheng: @pgm Is there really an issue with storing the values`1.0` vs `1`? I believe both the matrix and the tabular dataset validations coerces value types `continuous` to `Float64`. I personally liked standardizing all numbers to floats but let me know if there is a specific reason we should preserve original values

jessica-cheng: Nevermind

breadbox/tests/io/test_data_validation.py
```diff
@@ -137,3 +137,43 @@ def to_csv(*args, **kwargs):
...
+            "col4": ColumnMetadata(col_type=AnnotationType.categorical),
+            "col5": ColumnMetadata(col_type=AnnotationType.categorical),
+        },
+        "ID",
+    )
+    assert df["col1"].to_list() == [1.0, 2.1, 3.0]
+    assert df["col1"].dtype == pd.Float64Dtype()
+    assert df["col2"].to_list() == ["1", "2.1", "3"]
+    assert df["col2"].dtype == pd.StringDtype()
+    assert df["col3"].to_list() == ["1.0", "1.1", "1.0"]
```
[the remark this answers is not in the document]
pgm: re "Here we have a theoretical case where a column of type Categorical with values [1, 1.1, 1] is transformed to ["1.0", "1.1", "1.0"]" 

Yes, I'd call that a bug. Realistically, I don't think it's a common case that will get hit. Just like the original bug, I pointed out. I didn't think that that was a common case. I largely was advocating for fixing it now because I understood where it was coming from and thought the fix was going to be trivial. This makes me think that there is another step somewhere which is doing an implicit conversion, but finding it would take some debugging. I'd be fine, merging what you've got and just considering that as a open bug for the future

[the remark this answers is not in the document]
pgm: re: "Perhaps to address your concern, only columns that are expected to be of type text can be read with the specified dtype string?"

Could we have the reading of the CSV explicitly specify the types? I am largely not a fan of letting it implicitly guess the types of columns because there are data patterns, which will cause it to guess wrong. I thought sayings all columns were type string was going to be the simplest solution, but explicitly providing the types for each column is also a great solution. Just requires a little bit more work to construct the dictionary of column to type.

PR #184, opened 2025-01-30
feat(breadbox): Aggregate matrix dataset

jessica-cheng:
Add aggregation to api endpoint: `/datasets/matrix/{dataset_id}` which can return a subsetted matrix df as well as aggregate the df on either axis.

### DIFF
--- a/breadbox/breadbox/service/dataset.py
+++ b/breadbox/breadbox/service/dataset.py
@@ -173,12 +173,11 @@ def _aggregate_matrix_df(
     }
 
     # Replace None with numpy nan so np.nanpercentile works
-    df = df.replace({pd.NA: np.nan})
+    nan_df: pd.DataFrame = df.replace({pd.NA: np.nan})
 
-    df = df.agg(
+    return nan_df.agg(
         enum_to_agg_method[aggregation], axis=0 if aggregate_by == "samples" else 1
     )
-    return df
 
 
 def get_subsetted_tabular_dataset_df(
--- a/breadbox/breadbox/service/dataset.py
+++ b/breadbox/breadbox/service/dataset.py
@@ -6,6 +6,8 @@ from breadbox.schemas.dataset import (
     FeatureSampleIdentifier,
     MatrixDimensionsInfo,
     TabularDimensionsInfo,
+    AggregationMethod,
+    ValueType,
 )
 from breadbox.service import metadata as metadata_service
 import logging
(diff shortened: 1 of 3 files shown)


breadbox/tests/api/test_datasets.py
```diff
@@ -2489,6 +2489,138 @@ def test_get_matrix_dataset_data_by_labels(
...
+            json={"aggregate": {"aggregate_by": "samples", "aggregation": "25%tile"}},
+        )
+        """
+        Expected:
+        ------------------- 
+        A    1.0050
+        B    3.0025
+        C    2.0050
+        """
+        assert response.json() == {"A": 1.0050, "B": 3.0025, "C": 2.0050}
```
jessica-cheng, line 2560: Admittedly this is what `np.nanpercentile()` returns and I am just trusting the result. Please let me know if these results don't make sense!

rcreasi:
This looks great! And it looks like it was pretty easy to do 🙂

There's just one small thing that's different from the proposal. I had suggested the result set be wrapped:
```
{
  "mean": {
    "ACH-000001": 0.5815589549,
    "ACH-000002": 0.4823100117,
    "ACH-000004": 0.5686150194
  }
}
```
Just so the response has a consistent structure with or without aggregation (you either get a set of features or one "computed feature"). And you could imagine extending that by, say, passing a list of aggregations and getting them all back as separate "features". That would be neat! But we're probably never going to need that so I'm fine with keeping this as is.

breadbox/breadbox/service/dataset.py
```diff
@@ -143,9 +146,40 @@ def get_subsetted_matrix_dataset_df(
         )
         df = df.rename(index=label_by_id)
 
+    if dimensions_info.aggregate:
+        if dataset.value_type != ValueType.continuous:
+            raise UserError(
+                f"The `value_type` of '{dataset.name}' is '{dataset.value_type.value}'. Dataset must have continuous values! "
+            )
```
rcreasi, line 153: 💯

breadbox/tests/api/test_datasets.py
pgm: Looks reasonable to me.

pgm (approved):

jessica-cheng:
> This looks great! And it looks like it was pretty easy to do 🙂
> 
> There's just one small thing that's different from the proposal. I had suggested the result set be wrapped:
> 
> ```
> {
>   "mean": {
>     "ACH-000001": 0.5815589549,
>     "ACH-000002": 0.4823100117,
>     "ACH-000004": 0.5686150194
>   }
> }
> ```
> 
> Just so the response has a consistent structure with or without aggregation (you either get a set of features or one "computed feature"). And you could imagine extending that by, say, passing a list of aggregations and getting them all back as separate "features". That would be neat! But we're probably never going to need that so I'm fine with keeping this as is.

Oops! I missed that detail. I agree making the response look consistent makes sense to me. I'll change the response to include the aggregation method as well

PR #185, opened 2025-01-31
context selection is broken on Compound Sensitivity tab

rcreasi:
All the frontend code that relies on old DE2 APIs now has to explicitly opt-in. That will make it easy us for shutdown those endpoints and know what code needs to be updated. Unfortunately I missed out a lot of those places and now we're discovering them as bugs 🤦


snwessel (approved):
Looks good! Thanks Randy!

PR #186, opened 2025-02-03
show primary disease and lineage on hover in DE2

rcreasi:
hover


pgm (approved):
This is great, thanks

PR #187, opened 2025-02-03
Dataset form cleanup

jessica-cheng:
1. Hide "data type" field and default it to "User upload" when in "Simple" mode.
2. Add an example of the format of how a CSV file should be formatted before uploading.
3. Change group form field description

https://github.com/user-attachments/assets/f240e40a-da5a-49c8-9e54-0a6f6b37c6b9


pgm (approved):
LGTM

PR #188, opened 2025-02-04
Entity Summary breadbox support 

snwessel:
[Asana](https://app.asana.com/0/1165651979405609/1209255150160653)

**Goal**: update the compound sensitivity tab to work with breadbox datasets indexed by compound. 

**Background**: From what I learned poking around, it seems like this sensitivity tab is one of the older parts of our codebase - and was historically called "entity summary". This entity summary is also used on the Gene page's "Perturbation Effects" tab - so the component needs to be able to work with either genes or compounds. 

**Limitations:**
- I did my best to limit the scope of these changes - but it still ended up being a lot. There are a few legacy methods (like `get_compound_experiment_priority_sorted_datasets_with_compound`) which I rewrote with a better contract (`data_access.get_all_datasets_containing_compound` ), but I had to leave the old implementations in place because the tiles use them quite extensively. I marked the old implementations as "DEPRECATED" since they'll need to be removed before we can migrate compound data to breadbox.
- I have tested these changes to make sure that the pages still work with legacy datasets. However, I haven't tested them with an actual breadbox dataset yet because (as far as I know), we don't have any breadbox datasets indexed by compound yet. Once we generate one of those, I'll go back and debug this with real breadbox data. 

## Functionality Changes

**Before**:
* Graph axis is configured per-dataset
* Dropdown lets you select the dataset _and_ the compound experiment 
* Links to DE2 did some fancy logic to find a second dataset+feature for the compound 
* Screenshot:
image

**After my changes**:
* Axis scales to fit the data
* Dropdown only lets you select a dataset
* No compound experiment is displayed at the top of the page
* Links to DE2 open a similar plot in DE2 - a 1D density plot of whichever compound + dataset they are looking at, colored by lineage. This change let me delete a LOT of legacy functionality (all of `url_utils.py`)
* Screenshot:
image


portal-backend/depmap/compound/legacy_utils.py
```diff
@@ -1,35 +1,88 @@
...
+            experiments_by_compound_id[compound_id] = [comp_exp]
+            compounds_by_compound_id[compound_id] = compound
+
+    # Remove duplicate compound experiments, re-index by compound experiment
+    result = []
+    for compound_id, compound in compounds_by_compound_id.items():
+        experiments = experiments_by_compound_id[compound_id]
+        selected_experiment = sorted(experiments, key=lambda e: e.entity_id)[0]
+        result.append((selected_experiment, compound))
+    return result
```
snwessel, line 56: Refactored this method to use Phil's suggestion from [here](https://github.com/broadinstitute/depmap-portal/pull/175#discussion_r1930630207) (of collapsing by compound earlier in the process).

portal-backend/depmap/dataset/models.py
```diff
@@ -105,7 +105,7 @@ class Dataset(Model):
     __mapper_args__ = {"polymorphic_identity": "dataset", "polymorphic_on": type}
 
     @property
-    def nominal_range(self):
+    def nominal_range(self): # No longer used
```
pgm, line 108: I suggested that we get rid of this, but I think a cheap way we could add this back (one might say, the "right" way it should have been modeled actually) is determine the nominal range based on the *units* of the dataset.

Basically, all "gene effect" scores should be in the range something like (-1.1, 0.1), and all compound viability should be something like (0, 1) etc.

We'd probably have a little headache from how our units are written may not be consistent on different datasets, but I think determining the range from units makes more sense then needing to set it per-dataset.

Regardless, nothing to do here for now. Just mentioning it as a fyi

portal-backend/pyright-ratchet-errors.txt
```diff
@@ -584,11 +572,6 @@ test_views.py: error: Cannot access member "get_dose_enum" for type "DependencyD
 test_views.py: error: Cannot access member "label_without_compound_name" for type "CompoundDoseFactory"
 test_views.py: error: Cannot access member "label_without_compound_name" for type "CompoundDoseReplicateFactory"
 test_views.py: error: Object of type "None" is not subscriptable (reportOptionalSubscript)
-url_utils.py: error: Argument of type "Literal['ENABLED_FEATURES']" cannot be assigned to parameter "__s" of type "slice" in function "__getitem__"
```
pgm, line 587: More fixes!

portal-backend/tests/depmap/partials/entity_summary/test_models.py
```diff
@@ -145,7 +123,7 @@ def test_integrate_cell_line_information(empty_db_mock_downloads):
     cell_line_1 = CellLineFactory()
     cell_line_na = CellLineFactory()
     cell_line_3 = CellLineFactory()
-    gene = GeneFactory()
+    gene: Gene = GeneFactory() # pyright: ignore
```
pgm, line 126: So annoying, right??

pgm (approved):
Looks good. I do think it'll be good to test with BB data before merging in though. I'm wondering about looking things up by labels...

snwessel:
After trying things out with real breadbox data, I had to make a few more changes. The main issue was that the `compound_id` parameter that was being passed around in most of these legacy methods was actually an entity ID, which is a concept that doesn't exist in breadbox. The IDs breadbox uses actually map onto the `Compound.compound_id` field, not `Compound.entity_id`. 

**As a result, I changed a few more things:** 
1. I moved a few methods off of the `EntitySummary` class so I could have more control over how their called and what IDs are passed around. This also makes the code much easier to read (in my opinion) and continues moving us away from the "entity" models. 
2. Since I've made substantial changes to the `EntitySummary` class at this point, I also had to double check that it wasn't being used anywhere else. The "genomics association" tab had been using it, but is broken and being redesigned anyway, so I simplified/removed some of the backend for that and made sure the tab is always hidden.
3. When a compound's data is loaded, it is now being indexed by compound label. This means it will continue to be important for breadbox and the legacy backend to use the same labels. 
    - Sidenote: the dev data we use for compounds had lowercase labels (which doesn't match what we use in production). I updated it to match so we don't get errors when testing things locally. 
    
Once I have all the tests passing on this branch, I am planning to merge + deploy this to QA to double check that everything is working as expected 👍

PR #189, opened 2025-02-05
Add feature flag to not show dataset and group managers in public portal

jessica-cheng:


portal-backend/depmap/groups_manager/views.py
```diff
@@ -17,4 +14,6 @@ def view_groups_manager():
     """
     Entry point for groups manager
     """
+    if current_app.config["ENABLED_FEATURES"].is_public():
```
pgm, line 17: Unfortunately, this is not how the feature flags are intended to be used. We have a convention of creating flags that have names that try to make it obvious what functionality is enabled or disabled and test that. So we try not to have functionality directly call is public or any test on any environment name.

For example, here's a place where we are checking a feature flag in the TDA views:

```
@blueprint.route("/")
def view_tda_summary():
    if not current_app.config["ENABLED_FEATURES"].target_discovery_app:
        abort(404)

    return render_template("tda/index.html")
```

and then `target_discovery_app` is defined to return a value based on the environment.
```
    @property
    def target_discovery_app(self):
        return True
```

That one is perhaps not too interesting but here is a different property on feature flags:
```
    @property
    def access_control_and_private_resources(self):
        return self.is_skyros() or self.is_dmc_like()
```

Inside the property check for specific environments. This strategy is nice because it means that as long as we've defined, the future flag well and we've tested for it and all the places that are related to that functionality. It's easy to turn the functionality on or off for an environment by editing the definition of the property.

Ended up with this solution as a result of having a lot of trouble turning on functionality in environments because we had to search through the code base and find all of the places related to some functionality and update the check for the appropriate list of environments.

This allows us to have a single place we can change to enable or disable piece of functionality across the entire portal.

pgm: I think you could reuse the `access_control_and_private_resources` feature flag and just check that. That is capturing the same spirit of the functionality that were enabling. Any environments that we have access control, and private resources are the ones where we would want to show the data manager moving forward.

pgm (changes requested):

portal-backend/depmap/dataset_manager/views.py
```diff
@@ -14,4 +15,6 @@ def view_dataset_manager():
     """
     Entry point for dataset manager
     """
+    if current_app.config["ENABLED_FEATURES"].is_public():
```
pgm, line 18: see below

PR #190, opened 2025-02-05
Filter datasets by access

jessica-cheng:
In simple mode only, filter out datasets that user does not have write access to. In advanced mode, datasets in groups that user have read or write access are shown and are an option

### DIFF
--- a/frontend/packages/@depmap/dataset-manager/src/components/index.tsx
+++ b/frontend/packages/@depmap/dataset-manager/src/components/index.tsx
@@ -121,7 +121,7 @@ export default function Datasets() {
         let currentDatasets = await dapi.getBreadboxDatasets();
 
         if (!isAdvancedMode) {
-          const writeGroups = await getGroups();
+          const writeGroups = await dapi.getGroups(!isAdvancedMode);
           const group_ids = writeGroups.map((group) => {
             return group.id;
           });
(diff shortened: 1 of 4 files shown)


frontend/packages/@depmap/dataset-manager/src/components/index.tsx
```diff
@@ -53,7 +53,10 @@ export default function Datasets() {
       dapi.updateDimensionType(dimTypeName, dimTypeArgs),
     [dapi]
   );
-  const getGroups = useCallback(() => dapi.getGroups(), [dapi]);
+  const getGroups = useCallback(() => dapi.getGroups(!isAdvancedMode), [
+    dapi,
+    isAdvancedMode,
+  ]); // write access set to true if not advanced mode
```
pgm, line 59: I appreciate this comment!

frontend/packages/@depmap/dataset-manager/src/components/index.tsx
```diff
@@ -115,7 +118,17 @@ export default function Datasets() {
   useEffect(() => {
     (async () => {
       try {
-        const currentDatasets = await dapi.getBreadboxDatasets();
+        let currentDatasets = await dapi.getBreadboxDatasets();
+
+        if (!isAdvancedMode) {
```
pgm, line 123: It appears get groups is only called here. Can't you skip creating the callback and just call `dapi.getGroups(!isAdvancedMode)` here?

The way that the callback was a function of his advanced mode I was expecting it to be used in advanced mode and non-advanced mode, but it looks like the only way this ever gets used is if isAdvancedMode is set to false. 

So I think this all works it's just it would be easier to follow the logic if that was explicit. Having the call back created as a layer of indirection where I have to go back up to the top to see how it was defined.

pgm (approved):
LGTM

PR #191, opened 2025-02-10
Fix dataset manager delete errors

jessica-cheng:
### Fixes:
- Update dataset manager to remove deprecated DELETE method for dimension type and use new method
- Fix UI bug where failing to delete a dataset while having a dimension type selected shows the error for dimension type instead of dataset error
**Actions that caused UI Bug:**
1. Go to Advanced Mode
2. Select a dimension type
3. Select a dataset (that user doesn't have write access for such as a public dataset)
4. Click 'Delete Datasets' button 

**Fixed UI**
Screenshot 2025-02-10 at 4 45 57 PM

### DIFF
--- a/frontend/packages/@depmap/dataset-manager/src/components/index.tsx
+++ b/frontend/packages/@depmap/dataset-manager/src/components/index.tsx
@@ -372,7 +372,6 @@ export default function Datasets() {
   };
 
   const deleteDatasetButtonAction = async () => {
-    let isDeleted = false;
     const datasetIdsSet = new Set(selectedDatasetIds);
 
     await Promise.all(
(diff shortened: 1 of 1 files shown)


frontend/packages/@depmap/dataset-manager/src/components/index.tsx
```diff
@@ -401,17 +414,21 @@ export default function Datasets() {
           (dt) => dt.name === selectedDimensionType.name
         );
         if (dimensionType.axis === "feature") {
```
pgm, line 416: Since both paths of this`if` block are the same why not remove the `if` entirely? It seems simpler to me

pgm (approved):
Some small coding style suggestions but otherwise seems fine

frontend/packages/@depmap/dataset-manager/src/components/index.tsx
jessica-cheng: You're right! Thanks for catching this

frontend/packages/@depmap/dataset-manager/src/components/index.tsx
```diff
@@ -367,18 +374,24 @@ export default function Datasets() {
...
-    }
+
+    await Promise.all(
+      Array.from(datasetIdsSet).map((dataset_id) => {
+        return dapi.deleteDatasets(dataset_id);
+      })
+    )
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      .then((vals) => {
+        isDeleted = true;
```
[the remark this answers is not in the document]
jessica-cheng: After looking at this more, I think I am not using the promise resolver correctly. There shouldn't be a need for setting `isDeleted` at all since the `.then()` method should execute when the promise resolves.  And since I'm using `Promise.all()` it will reject if _any_ of the calls fail and resolve when _all_ promises resolve. 

To summarize, what's happening here is once all datasets are successfully deleted, then I update the React state for the list of datasets and dimension types with dataset counts.

PR #192, opened 2025-02-11
switched dep mgmt to poetry and added packed cor module

pgm:
This contains a few related changes around building the pipeline docker image

1. Changed the way that python packages were installed to use poetry so that module versions are now pinned.
2. No longer creates a virtual environment for the main python environment because isolation isn't really necessary.
3. Removed installation of boto because we removed all references to AWS, and I can't think of anything else that would require it.
4. Added a new python dependency for installation of the packed correlation tables module, which is used for new correlation table format


naquib314 (approved):
LGTM

PR #194, opened 2025-02-12
Fix parsing unknown ids and styling bugs

jessica-cheng:
### Fixes

- Uploading datasets with unknown IDs caused page to crash. This is because the data structure for unknown IDs has changed and was no longer able to be parsed
- List of unknown IDs could be very long and therefore displaying it all on the modal means the modal can get extremely long. Additional logic to truncate list and styling for long strings added
Screenshot 2025-02-12 at 10 31 29 AM


pgm (approved):
Sure, sounds fine to me

PR #195, opened 2025-02-20
Fix celligner nan bug

alimourey:
These changes fix a bug that caused the k neighbors text input to get stuck on NaN if the user deleted everything in the input box.

The bug was caused by parse int of an empty string. This is fixed by using type="number" and min="0".

These changes also allow export of the nearest neighbors table data.


pgm (approved):
LGTM

PR #197, opened 2025-02-25
Removing unused API endpoints (Vector Catalog) 🎉 

snwessel:
[Asana](https://app.asana.com/0/1165651979405609/1209205317488196)

**Removing unused endpoints:**
* `vector_catalog`
  * `/data/catalog/children`
  * `/data/catalog/path`
  * `/data/vector/<path:id>`
*  `gene/has_rephub/<gene_symbol>`: this is unrelated to vector catalog, but I noticed it references to compound experiments and is not being used, so I think it makes sense to remove. 

Also removing any now-unused code  in the `vector_catalog` directory 🎉 However, I did have to leave a few things in that directory, including:
* the `SliceSerializer` which is used for parsing our legacy slice format
* a few static methods on the `InteractiveTree` class which are still used in a few places. 

Eventually, we'll move away from using those as well.  

**Remaining work**: There are some other endpionts under `/interactive` which are likely not being used anymore (they had been used by data explorer 1). I am going to make a follow-up task to review and remove those.


pgm (approved):
LGTM

PR #199, opened 2025-02-27
Filter metadata datasets

jessica-cheng:
UI Bug Fix: Filter dataset metadata options in dimension type edit form to include only tabular datasets with the dimension type to edit


pgm (approved):
LGTM

PR #200, opened 2025-03-03
feat(breadbox): added support for fetching cas values from legacy storage

pgm:
The legacy backend stores CAS values in a Google bucket. Breadbox stores CAS values in a sqlite database to minimize external dependencies.

In order to transition we're going to need to copy all the key value pairs from the bucket into the appropriate database, however, in the meantime, this set of changes will provide, graceful support for keys in both places.

These changes can be reverted after all keys have been copied. Since these changes are pretty minimal and going to go away eventually anyway I've just manually tested them as opposed to writing a test suite covering them.

To enable this behavior for development add `LEGACY_CAS_BUCKET = "depmap-dev-portal-cas"` to your .env file. If this setting is left out, then these changes have no effect.


rcreasi (approved):
This looks good to me. Pretty clever solution!

PR #201, opened 2025-03-04
feat(breadbox): changes to pre-compute correlation tables to support new "correlation analysis" module

pgm:
I started tweaking our existing process for precomputing correlations in order to support a correlation module that mirrors that which you find on the PRISM portal. 

There were some small tweaks to the methodology (ie: thresholds on min to keep for both positive and negative correlations) needed and ultimately I copied correlation.py and created a new script and set of conseq rules. This allows us to run the two in parallel and get two independent set of results. My plan is to get the pipeline portion working, and then the breadbox loader changes needs to import these tables. Lastly will be adding the frontend. Only after all parts are working, can we drop the old correlation tables and switch to using this new version entirely.


pipeline/cor_analysis/create_cor_analysis_pairs.py
```diff
@@ -0,0 +1,52 @@
...
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("in_filename")
+    parser.add_argument("out_filename")
+    args = parser.parse_args()
+
+    with open(args.in_filename, "rt") as fd:
+        inputs = json.load(fd)
+        print("warning: ignoring input: ", inputs)
+        # todo change this so we take the taiga IDs from inputs
```
naquib314, line 17: Maybe this will change after the todo is done. But for now it seems a bit confusing to me that saying "Ignoring input" but then using those inputs unless I am missing something.

pipeline/cor_analysis/cor_analysis.conseq
```diff
@@ -0,0 +1,69 @@
+
+rule create_cor_analysis_pairs:
+    inputs:
+        a_set=all {'type': 'cor-analysis-a'},
+        b_set=all {'type': 'cor-analysis-b'},
+        script=fileref("create_cor_analysis_pairs.py")
+    
+    run "bash" with """
```
naquib314, line 8: I had no idea that you could do this.

packed-cor-tables/packed_cor_tables/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+    for f_givenid, cbuf in cursor.fetchall():
+        df = _unpack(cbuf)
+        df["feature_given_id_0"] = f_givenid
+        df["feature_given_id_1"] = _map_dim_index_to_given_ids(cursor, 1, df["dim_1"])
+        df.drop(columns=["dim_1"], inplace=True)
+        dfs.append(df)
+    return pd.concat(dfs, ignore_index=True)
+
+
+def _unpack(cbuf):
```
naquib314, line 103: ❤️

pipeline/cor_analysis/create_cor_analysis_pairs.py
pgm: Whoops! I should take that out. I had hardcoded the value of `inputs` at one point. I'll fix.

pipeline/cor_analysis/cor_analysis.conseq
rcreasi: My conseq syntax highlighter will even switch languages to bash :)

PR #202, opened 2025-03-04
Context Explorer v2 changes

acameron13:
- Adding subtype tree and context matrix creation to the data-prep-pipeline
- Editing the get_context_analysis script to return results for all new contexts


data-prep-pipeline/scripts/subtype_tree/create_context_matrix.py
```diff
@@ -0,0 +1,244 @@
...
+    parser.add_argument("output", help="filepath to write the output")
+    args = parser.parse_args()
+
+    context_matrix = create_context_matrix(
+        args.model,
+        args.molecular_subtypes,
+        args.subtype_tree
+    )
+
+    if context_matrix is not None:
```
alimourey, line 241: Is there ever a case where the context matrix being None isn't a gigantic problem? If None is not an acceptable result, should we just let this blow up here so we're aware sooner?

data-prep-pipeline/scripts/subtype_tree/create_subtype_tree.py
```diff
@@ -0,0 +1,619 @@
...
+            of the subtype tree. In this case, it assumes that all Level0 nodes have
+            been properly added
+
+    Outputs:
+        - mst_tree (pandas df): An extended molecular subtype tree with this gene's
+            sub-tree added in 
+    '''
+    
+    #sorting by full subtype name is important so that the level and parent
+    #assignments don't get jumbled
```
alimourey, line 372: I don't think I understand what this comment means. Do you have an example of what the level and parent assignments getting jumbled looks like?

data-prep-pipeline/scripts/subtype_tree/create_subtype_tree.py
```diff
@@ -0,0 +1,619 @@
...
+    subtype_tree = create_subtype_tree(
+        args.model,
+        args.oncotree,
+        args.molecular_subtypes,
+        args.genetic_subtypes_whitelist_folder,
+        args.genetic_subtypes_whitelist_filename
+    )
+
+    if subtype_tree is not None:
+        subtype_tree.to_csv(args.output)
```
alimourey, line 617: Like above, I wonder if we should take out this None check. Maybe I'm misunderstanding how this works.

data-prep-pipeline/scripts/subtype_tree/create_subtype_tree.py
acameron13: Just added a more in-depth explanation of what I mean!

data-prep-pipeline/scripts/subtype_tree/create_context_matrix.py
```diff
@@ -0,0 +1,241 @@
+import argparse
+import pandas as pd
+from taigapy import create_taiga_client_v3
+
+# TODO FOR 25Q2: REMOVE THE USE OF THIS TEMPORARY MODEL FILE
+temp_model_id = 'alison-test-649a.18/Model_temp_between_24q4_25q2'
```
pgm, line 6: @acameron13 , is this supposed to be here still?

This gives a vibe of "There's a change in 25Q2 which we don't have in the 24Q4 Model.csv, so use this temporary one which has the change"

If I'm correct, I'd say this is the sort of thing that often gets forgotten about in the shuffle of a release, especially when there's a big gap in time and people between the person who knows what's going on here (you today) and the person who will be doing the release (someone else in a few months)

There are a few things I typically do for situations like this:
1. Create an asana task describing the change that needs to be made when the release is in progress and assign it to someone with an approximate date so it doesn't get lost.
2. Some defensive coding: leave an assertion in place to detect whether this hack is necessary and throw an error you can tell we're not running it on the 24Q4 release. This could be as simple as adding a statement like the following would be sufficient.

```
assert model_taiga_id == 'depmap-24q4-somethingsomething/Model", "If this has changed see asana task XXXXXX for instructions how this should be updated and why"
```

data-prep-pipeline/scripts/subtype_tree/create_context_matrix.py
```diff
@@ -0,0 +1,241 @@
...
+        st_ctx.loc[st_models] = 1
+
+        #save the series to the list of columns
+        ctx_cols.append(st_ctx)
+
+    context_matrix = pd.DataFrame(ctx_cols).T
+
+    return context_matrix
+
+def sanity_check_results(subtype_tree, context_matrix):
```
pgm, line 182: I appreciate this. ❤️

data-prep-pipeline/scripts/subtype_tree/create_subtype_tree.py
```diff
@@ -0,0 +1,618 @@
+import argparse
+import re
+import itertools
+import numpy as np
+import pandas as pd
+from taigapy import create_taiga_client_v3
+
+# TODO FOR 25Q2: REMOVE THE USE OF THIS TEMPORARY MODEL FILE
+temp_model_id = 'alison-test-649a.18/Model_temp_between_24q4_25q2'
```
pgm, line 9: oh boy, this in a few places, huh? Definitely recommend asserts so we don't miss changing any

data-prep-pipeline/scripts/subtype_tree/create_subtype_tree.py
```diff
@@ -0,0 +1,618 @@
...
+
+    #copy the parent, and add all of the new information
+    new_node = parent_node.copy()
+
+    #reset all node meta fields
+    new_node['NodeLevel'] = parent_node['NodeLevel'] + 1
+    new_node['NodeName'] = new_name
+    new_node['NodeSource'] = new_source
+    new_node[code_col] = new_code
+    new_node['OncotreeCode'] = np.nan #if we are adding a custom node, there is no existing Oncotree code
```
pgm, line 170: minor point: Originally pandas used NaN to represent missing data. This has resulted in a whole lot of headaches and I think the authors and community now have accepted that was a bad idea. To remedy this, at some point pandas introduced new data types which are aware of a special NA value. 

When writing new code, I recommend using pandas's NA instead of NaN because it can avoid confusion/problems. In this context, it just is a matter of changing this line to

```
new_node['OncotreeCode'] = pd.na
```

pgm (commented):
Generally looks fine, but I strongly recommend adding some asserts to detect when we change versions of the model file.

I do have a question though: This looks like it's changing the script that generates files for context explorer. @alimourey , these changes are needed for context explorer v2, but would break context explorer v1, correct? Or are they safe to use with context explorer v1?

(Concretely, I'm wondering if these changes should be merged into `master` or into the branch that Ali is using for her development.)

naquib314 (approved):
Have one minor comment. Apart from that lgtm.

alimourey:
@acameron13 context-explorer-v2 is now up to date with master. You probably also need to merge master into this branch.

PR #204, opened 2025-03-05
Disable add dataset button

jessica-cheng:
- Disable button to add dataset if user does not belong in any groups with write access. 
- Refactored by passing groups as props instead of passing the `getGroups()` callback

Screenshot 2025-03-05 at 4 37 05 PM


pgm (approved):
LGTM

PR #207, opened 2025-03-06
fix(breadbox): Only allow public datasets for dimension type metadata

jessica-cheng:
Only allow public tabular datasets to be associated with dimension types

### DIFF
--- a/frontend/packages/@depmap/dataset-manager/src/components/DimensionTypeForm.tsx
+++ b/frontend/packages/@depmap/dataset-manager/src/components/DimensionTypeForm.tsx
@@ -34,7 +34,7 @@ export default function DimensionTypeForm(props: DimensionTypeFormProps) {
         (d) => {
           return (
             d.index_type_name === dimensionTypeToEdit.name &&
-            d.group.name === "Public"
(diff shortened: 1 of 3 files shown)


frontend/packages/@depmap/dataset-manager/src/components/DimensionTypeForm.tsx
```diff
@@ -31,9 +30,12 @@ export default function DimensionTypeForm(props: DimensionTypeFormProps) {
 
   React.useEffect(() => {
     if (isEditMode && dimensionTypeToEdit) {
-      const datasetsWithDimensionType: TabularDataset[] = datasets.filter(
+      const publicDatasetsWithDimensionType: TabularDataset[] = datasets.filter(
         (d) => {
-          return d.index_type_name === dimensionTypeToEdit.name;
+          return (
+            d.index_type_name === dimensionTypeToEdit.name &&
+            d.group.name === "Public"
```
pgm, line 37: If we're going to hardcode something, I think it should be the ID not the label. In the backend we have a constant for the public group ID for that purpose

pgm (approved):
only change I'd recommend is matching by ID instead of name

PR #210, opened 2025-03-06
Add submission message details in edit dataset form

jessica-cheng:
Before, no info was relayed to user on whether an edit to a dataset was successful or not. This change adds a status message on submitting the form.

The below image is a POC to show that an error message is shown when a dataset edit fails 
Screenshot 2025-03-06 at 4 02 27 PM


pgm (approved):
LGTM

PR #212, opened 2025-03-06
Update DE2 in Elara to work with Breadbox APIs

rcreasi:
This updates the version of Data Explorer 2 visible in Elara to use only Breadbox APIs (i.e. it doesn't rely on the legacy Portal backend at all).

Any calculations or data processing (beyond just fetching raw data) have been moved to the frontend. In order to make minimal changes to the existing UI code, these things still appear to happen as part of the network request. Some operations with names like "fetchXYZ" are actually doing quite a bit more than fetching.

There are now two sets of sets of API methods (implemented as React contexts):
- DataExplorerApi
  - This is only referenced in code specific to the new version (none of its methods are implemented in the legacy version). Its methods fit well into the Breadbox model and have a sensible contract that can hang around after we sunset the DE2 backend in the legacy Portal.

- DeprecatedDataExplorerApi
  - These functions exist to keep the legacy version of DE2 running. Some are also used in the new Elara version but only to cause minimal disruption to the code as described above. Things like fetchCorrelation() are fetching raw data and then performing in JavaScript calculations the backend used to do. I plan to rename/rework things to make that more obvious, once there is no longer a legacy version to support.

While this commit represents a fairly feature-rich version of DE2, the following are not yet implemented:
- Pre-computed associations
- Dataset details modal
- "Set selection from a context" button
- "Download" button in Context Manager
- Always showing Lineage and Primary Disease labels on all points that are models


pgm:
Hurray!

PR #214, opened 2025-03-10
breadbox(perf): Small breadbox performance tweak

snwessel:
This may no longer be relevant, since it sounds like Randy found a different solution that doesn't use this parameter - but I wanted to commit my changes anyway since they improve performance a bit. 

**Background**: Randy noticed that this endpoint takes a long time when `?show_only_dimensions_in_datasets=true` is set (for example, [here](https://dev.cds.team/depmap-iqa/breadbox/types/dimensions/gene/identifiers?show_only_dimensions_in_datasets=true))

**Solution**: moving a couple of things into the SQLAlchemy query, so that only distinct given IDs are loaded into memory (instead of the full `Dimension` objects)

This makes the endpoint about 4x faster with my somewhat-realistic local data.

### DIFF
--- a/breadbox/breadbox/crud/dataset.py
+++ b/breadbox/breadbox/crud/dataset.py
@@ -6,6 +6,7 @@ import warnings
 
 import pandas as pd
 from sqlalchemy import and_, func, or_
+from sqlalchemy.sql import distinct
 from sqlalchemy.orm import aliased, with_polymorphic
 
 from breadbox.db.session import SessionWithUser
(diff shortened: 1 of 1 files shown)


breadbox/breadbox/crud/dataset.py
```diff
@@ -783,19 +784,21 @@ def get_unique_dimension_ids_from_datasets(
     unique_dims = set()
```
jessica-cheng, line 784: Looks like we can remove this with your changes!

breadbox/breadbox/crud/dataset.py
```diff
@@ -805,12 +808,8 @@ def get_unique_dimension_ids_from_datasets(
...
         .all()
-    )
+    }
     # Combine dimension type's dimension given ids from datasets
-    for m_dim in matrix_dimensions:
-        unique_dims.add(m_dim.given_id)
-
-    for t_dim in tabular_dimension_ids:
-        unique_dims.add(t_dim.dimension_given_id)
+    unique_dims = tabular_dimension_ids.union(matrix_dimension_ids)
```
jessica-cheng, line 813: 👍

jessica-cheng (approved):
LGTM!

PR #217, opened 2025-03-13
feat(breadbox): Support list of strings values matrix

jessica-cheng:


breadbox/breadbox/io/data_validation.py
```diff
@@ -134,6 +134,26 @@ def _validate_dimension_type_metadata_file(
     return df
 
 
+def _parse_list_strings(val):
+    try:
+        deserialized_str_list = json.loads(val)
```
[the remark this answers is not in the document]
pgm: Yeah, this is a question that doesn't have an obvious answer based on prior art.

Programming languages typically have [] and a "nil" like type as distinct concepts and types. However, SQL databases have historically blurred the concept. (ie: in SQL "" and null are typically equivalent)

With my programming background and my biases towards well typed systems, I like the idea of having elements be either arrays _or_ null, which is to say `[] != nil`.

I also think it's more expressive to have the two types be different. If nil is used to model holes in the dataframe, that means you can distinquish between "this gene and cell line have no protein changes" (the value is known) and "this gene and cell line did not have protein changes recorded for it" (the value is unknown)

When you say, "allowing empty lists" do you mean that's what you chose as well?

breadbox/breadbox/io/data_validation.py
```diff
@@ -134,6 +134,26 @@ def _validate_dimension_type_metadata_file(
...
+    except Exception as e:
+        raise FileValidationError(
+            f"Value: {val} must be able to be deserialized into a list. Please make sure values for columns of type list_strings are a stringified list (ex: {example_list_string})"
+        ) from e
```
pgm, line 144: ❤️ Loving the chained exceptions so the original failure isn't lost

breadbox/breadbox/io/data_validation.py
```diff
@@ -134,6 +134,26 @@ def _validate_dimension_type_metadata_file(
...
+
+    if not isinstance(deserialized_str_list, list):
+        raise FileValidationError(
+            f"Value: {val} must be able to be deserialized into a list. Please make sure values for columns of type list_strings are a stringified list (ex: {example_list_string})"
```
pgm, line 148: This error message references `example_list_string` which appears to be defined below, so I'd expect an "unassigned variable" error here. (I'd expect pyright to detect this actually. Is `example_list_string` defined somewhere else as well?

breadbox/breadbox/io/data_validation.py
```diff
@@ -170,6 +190,18 @@ def _validate_data_value_type(
...
         return int_df
+    elif value_type == ValueType.list_strings:
+
+        def validate_list_strings(val):
+            if not pd.isnull(val):
+                _parse_list_strings(val)
+                return val
+            else:
+                # hdf5 will stringify 'None' or '<NA>'. Use empty string to represent NAs instead
+                return ""
```
pgm, line 201: Great, love the comment

breadbox/breadbox/io/filestore_crud.py
```diff
@@ -11,13 +12,21 @@
...
+    filestore_location: str,
 ):
     base_path = os.path.join(filestore_location, dataset_id)
     os.makedirs(base_path)
 
+    if value_type == ValueType.list_strings:
+        dtype = "str"
+    else:
+        dtype = "float"
+
```
jessica-cheng, line 27: I'm aware I can probably directly pass `ValueType` to `write_hdf5_file()` instead but I feel like the file it is in `hdf5_utils.py` attempts to decouple as much Breadbox related names from hdf5 utility functions. Or maybe I'm reading too much into this 😅

pgm (approved):
Looks great!

breadbox/breadbox/io/data_validation.py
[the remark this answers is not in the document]
pgm: Actually, I can tell from the code, that's how it's written. Looks good to me!

breadbox/breadbox/io/filestore_crud.py
```diff
@@ -94,6 +103,10 @@ def get_df_by_value_type(
         # Convert numerical values back to origincal categorical value
         df = df.astype(int)
         df = df.applymap(lambda x: dataset_allowed_values[x])
+    elif value_type == ValueType.list_strings:
+        # NOTE: String data in HDF5 datasets is read as bytes by default
```
jessica-cheng, line 107: I decided there was no need to first read out the hdf5 as string. This would be more memory efficient anyways

breadbox/breadbox/io/data_validation.py
jessica-cheng: Yes I fixed this error. I missed the earlier references while refactoring (oops)

breadbox/tests/api/test_dataset_uploads.py
```diff
@@ -224,6 +224,70 @@ def mock_return_failed_task(result):
...
+            },
+            headers=headers,
+        )
+        assert_status_ok(matrix_subset)
+        matrix_subset_result = matrix_subset.json()
+        assert matrix_subset_result == {
+            "A": {"ACH-1": 1.01, "ACH-2": None},
+            "B": {"ACH-1": 2, "ACH-2": 3.0},
+            "C": {"ACH-1": None, "ACH-2": None},
+        }
```
jessica-cheng, line 289: I couldn't immediately find a test that covers whether empty continuous values were read out correctly so added one just for my own sanity.

PR #220, opened 2025-03-18
Handling OncRef privacy based on deploy environment

acameron13:
Also in this PR:
- making sure that the model id's in the data availability matrix exactly match the model id'd in the context matrix
- some minor adjustments to how files are saved

### DIFF
--- a/pipeline/context_explorer/get_context_analysis.py
+++ b/pipeline/context_explorer/get_context_analysis.py
@@ -95,28 +95,31 @@ def load_oncref_data(tc, oncref_auc_taiga_id):
     return auc_matrix, log_auc_matrix
 
 
-def format_selectivity_vals(repurposing_table, oncref_table):
+def format_selectivity_vals(repurposing_table_path, oncref_table_path):
+    selectivity_dfs = []
 
     # reformat repurposing compounds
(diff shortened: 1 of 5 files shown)


pipeline/context_explorer/get_context_analysis.py
```diff
@@ -132,41 +135,51 @@ def load_all_data(
...
-        "context_matrix": context_matrix,
-        "gene_effect": gene_effect,
-        "gene_dependency": gene_dependency,
-        "rep_sensitivity": rep_sensitivity,
-        "oncref_log_aucs": oncref_log_aucs,
-        "oncref_aucs": oncref_aucs,
-        "selectivity_vals": selectivity_vals,
-    }
+    all_data_dict["datasets_to_test"] = datasets_to_test
+    all_data_dict["data_for_extra_cols"] = data_for_extra_cols
```
alimourey, line 180: I'm a little confused about what "data_for_extra_cols" is supposed to be doing. Is this something that was needed to handle Oncref being None for public?

alimourey (approved):
Looks good! I was confused about one of your changes, but overall I think it's good and can be merged in when you're ready so we can test it out end to end

PR #221, opened 2025-03-20
perf(breadbox): Sparse matrix chunked storage

jessica-cheng:
Large string matrices take up a lot of space when stored as an hdf5 file (2k rows x 18 columns resulted in over 1GB!). Our only use case so far where we store a list of strings matrix as strings is a pretty sparse matrix. So instead of writing the entire matrix into hdf5, we only store nonnull values and add chunked storage for more optimal storage size of the hdf5 file. This heuristic is defined for a sparse matrix of any dtype, not just string hdf5 datasets.

### DIFF
--- a/breadbox/breadbox/io/hdf5_utils.py
+++ b/breadbox/breadbox/io/hdf5_utils.py
@@ -24,32 +24,35 @@ def create_index_dataset(f: h5py.File, key: str, idx: pd.Index):
 def write_hdf5_file(path: str, df: pd.DataFrame, dtype: Literal["float", "str"]):
     f = h5py.File(path, mode="w")
     try:
-        dataset = f.create_dataset(
-            "data",
-            shape=df.shape,
-            dtype=h5py.string_dtype() if dtype == "str" else np.float64,
-            # data=df.values,
-            chunks=(1, 1),
-        )
-        rows, cols = np.where(df.notnull())
-        for row, col in zip(rows, cols):
-            dataset[row, col] = df.iloc[row, col]
+        # Get the row,col positions where df values are not null
+        rows_idx, cols_idx = np.where(df.notnull())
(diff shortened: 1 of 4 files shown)


breadbox/breadbox/io/hdf5_utils.py
```diff
@@ -24,12 +24,35 @@ def create_index_dataset(f: h5py.File, key: str, idx: pd.Index):
...
+            )
+            # only insert nonnull values into hdf5 at given positions
+            for row_idx, col_idx in zip(rows_idx, cols_idx):
+                dataset[row_idx, col_idx] = df.iloc[row_idx, col_idx]
+        else:
+            if dtype == "str":
+                # NOTE: hdf5 will fail to stringify None or <NA>. Use empty string to represent NAs instead
+                df = df.fillna("")
+
+            # NOTE: For a large and dense string matrix, the size of the hdf5 will be very large. Right now, list of string matrices are a very rare use case and it is unlikely we'll encounter one that is not sparse. However, if that changes, we should consider other hdf5 size optimization methods such as compression
```
jessica-cheng, line 49: @pgm So far we only have one use case for storing string data in hdf5 which happens to be a sparse matrix. We've only addressed how to handle sparse string/list of string matrices. Should we wait until this becomes a more common use case before addressing how to handle denser list of string matrices?

pgm: The code as written looks like it should handle dense str matrices, correct?

Is there something that it's missing or you're concerned about?

pgm: The only thing I think we'd want (for completeness) is a unit test to verify that the other code path works with strings, but on first glance it looks fine.

pgm (approved):

PR #223, opened 2025-03-24
perf(breadbox): Performance improvements for Elara

snwessel:
[Asana](https://app.asana.com/0/1165651979405609/1209671857588613/f)

**Several relatively low-effort performance improvements in breadbox, including:**
* Improving the way we load metadata for a specific dataset - now using a subquery instead of loading two separate results from large SQL queries. Note: I know subqueries are still slow, but they're easy to read, and this change made this particular operation more than 5x faster - it's now taking a negligable amount of time for even with 2-3k results. 
* Adding a method `get_matrix_dataset_given_ids` which can be used instead of `get_matrix_dataset_features` or `get_matrix_dataset_samples`, when we don't need those full objects. This makes a moderate performance improvement to a bunch of different breadbox endpoints. 
* Adding an optional `limit` parameter to the /identifiers endpoint. This was easy to add and seems like something that could be useful when initially populating the dropdown. We may not need all 2-3k results if we're going to send another request anyway once the user starts typing. 
* Using `TypedDict` instead of Pydantic's `BaseModel` for endpoints that return a long list of simple objects 

**Before the changes, loading this page in the test environment consistently took 5.9-6.8 seconds** (URL [here](https://dev.cds.team/depmap-test-perf/breadbox/elara/?p=eJxVjkFrwzAMhf-Lzg20tdumgR523m07jmG0SAmGxDG2aV2C_3vlQjd2EZLe96S3gp-WZNLdM3RA7KJNd7Mj2IB1xPlP8TN6My_Ek2hk54ouLkK3Qq4Fs40vOuDNxMn2LCiOY-ARk8CiDDbEJNt-cYlzqsbfWy_3yK4aHc51ev94-5SJsw-Vvlyg-1rhiqGS9srOWIKyAdXqA3xLI1JNVaQlTBg5VaIDfSJ1PG9Vszvth0YfjtS0rLhpFWpN_Q-hHuTRM_e_KKWUBz-KYTk) (reload with hard refresh)):
image

**After the changes, the page takes about 4.2-4.6 seconds and the id/label loading endpoints are now significantly faster:** 
image

**Future performance improvement suggestions:**
Seems like the next easiest performance improvement would be in frontend changes:
* Randy had mentioned that (when he has time), he's going to make an update to only call the `/identifiers` endpoint once. This will shave off another ~0.6 seconds.  
* Putting a limit on that initial call to `/identifiers` will also speed this up substantially - since most of the time is spent on serializing this large result. Once the user starts typing, another request is made anyway to the dimension search endpoint.

### DIFF
--- a/breadbox/breadbox/compute/dataset_tasks.py
+++ b/breadbox/breadbox/compute/dataset_tasks.py
@@ -2,10 +2,11 @@ from dataclasses import dataclass
 import dataclasses
 from io import BytesIO
 import time
-from typing import Any, Callable, Dict, List, Optional, Set, TypedDict, Union, Literal
+from typing import Any, Callable, Dict, List, Optional, Set, Union, Literal
 from logging import getLogger
 from uuid import UUID, uuid4
 from breadbox.io.upload_utils import create_upload_file
+from typing_extensions import TypedDict
 
 import celery
 
--- a/breadbox/breadbox/crud/dataset.py
+++ b/breadbox/breadbox/crud/dataset.py
@@ -462,7 +462,7 @@ def get_matrix_dataset_features(
     return dataset_features
 
 
-def get_matrix_dataset_samples( # TODO: replace this?
+def get_matrix_dataset_samples( 
     db: SessionWithUser, dataset: MatrixDataset
 ) -> list[DatasetSample]:
     assert_user_has_access_to_dataset(dataset, db.user)
@@ -791,8 +791,6 @@ def get_subset_of_tabular_data_as_df(
 
     return pivot_df["value"]
 
-
-from time import perf_counter 
 
 def get_unique_dimension_ids_from_datasets(
     db: SessionWithUser, dataset_ids: List[str], dimension_type: DimensionType
--- a/breadbox/breadbox/crud/dataset.py
+++ b/breadbox/breadbox/crud/dataset.py
@@ -476,6 +476,7 @@ def get_matrix_dataset_samples( # TODO: replace this?
 
     return dataset_samples
 
+
 def get_matrix_dataset_given_ids(
     db: SessionWithUser, dataset: Dataset, axis: str
 ) -> List[str]:
@@ -799,7 +800,6 @@ def get_unique_dimension_ids_from_datasets(
     """
     Returns a unique set of dimension given ids from matrix and tabular datasets based on the given dimension type
     """
-    start = perf_counter()
     if dimension_type.axis == "feature":
         matrix_dimension_class = DatasetFeature
     else:
@@ -817,7 +817,6 @@ def get_unique_dimension_ids_from_datasets(
         )
         .all()
     }
-    print(f"Time taken to get MATRIX dimension ids: {perf_counter() - start}") # below this point is neglegible
     # Get all tabular identifiers for that dimension type
     tabular_given_ids = {
         given_id for (given_id,) in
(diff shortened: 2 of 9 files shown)


breadbox/breadbox/compute/dataset_tasks.py
```diff
@@ -2,10 +2,11 @@
 import dataclasses
 from io import BytesIO
 import time
-from typing import Any, Callable, Dict, List, Optional, Set, TypedDict, Union, Literal
+from typing import Any, Callable, Dict, List, Optional, Set, Union, Literal
 from logging import getLogger
 from uuid import UUID, uuid4
 from breadbox.io.upload_utils import create_upload_file
+from typing_extensions import TypedDict
```
snwessel, line 9: Pydantic gives a warning if you try to import `TypedDict` from `typing` on a python version < 3.12 - this is fixing that warning

breadbox/breadbox/crud/dataset.py
```diff
@@ -794,19 +816,60 @@ def get_unique_dimension_ids_from_datasets(
...
+    metadata_vals_by_id = (
+        db.query(TabularColumn)
+        .filter(
+            and_(
+                TabularColumn.dataset_id == dimension_type.dataset_id,
+                TabularColumn.given_id == metadata_col_name,
+            )
+        )
+        .join(TabularCell)
+        .filter(TabularCell.dimension_given_id.in_(given_id_subquery))
```
snwessel, line 871: I don't feel 100% about this pattern of filtering by `._in(subquery)` - which has the potential to be slow with very large numbers of IDs. However, it's easy to read, and this change made this particular operation more than 5x faster - it's now taking a negligable amount of time (0.08 seconds) for even with 2-3k given IDs.

If we do this operation on any datasets that are an order of magnitude larger, it might be worth revisiting.

breadbox/breadbox/crud/dataset.py
```diff
@@ -477,6 +477,28 @@ def get_matrix_dataset_samples(
...
+    assert_user_has_access_to_dataset(dataset, db.user)
+
+    if axis == "feature":
+        dimension_class = DatasetFeature
+    elif axis == "sample":
+        dimension_class = DatasetSample
+    else:
+        raise ValueError(f"Invalid axis: {axis}")
+
+    given_ids = (
```
pgm, line 492: If this is a method worth improving, we could easily make it more efficient by create an index on dimension_class (dataset_id, given_id)

breadbox/breadbox/crud/dataset.py
pgm: Why do you think `.in(subquery)` is slow?

pgm: You could also write this with an `exists` clause instead of `in` clause would give you similar readability but I expect the performance to be comparable. (Modulo the quality of the query optimizer)

breadbox/breadbox/crud/dimension_types.py
```diff
@@ -277,14 +277,15 @@ def get_dimension_type_metadata_col(
...
-        db.query(TabularCell)
-        .join(TabularColumn)
+        db.query(TabularColumn)
         .filter(
             and_(
                 TabularColumn.dataset_id == dimension_type.dataset_id,
                 TabularColumn.given_id == col_name,
             )
         )
+        .join(TabularCell)
```
pgm, line 287: just curious: is the move of this join down a few lines, a cosmetic change or a performance change?

pgm (approved):
Looks fine to me

breadbox/breadbox/crud/dimension_types.py
snwessel: My thinking was that filtering before doing the join would make things faster because there's a much smaller total number of records involved in the join. For example, if there are 20 columns of metadata and 2000 rows, but we only care about the "label" column, this shrinks your join (cross product) to 2000x1=2000 instead of 2000*20=40000.

In this case, it probably doesn't make a meaningful impact though, because I don't think the join is actually the slow part 🤷 I did a little profiling of this later and realized that the `fetch_all` operation (loading the query results) was taking something like 98% of the time this query takes to execute 😓

breadbox/breadbox/crud/dataset.py
snwessel: Actually yeah, thinking a bit more about this, it's really not an issue in this case. I have a placeholder in my mind that subqueries are slow and/or bad practice. In reality, it's really just correlated subqueries that are slow - where you use a value from the parent query, which forces the subquery to be re-executed every time its used (not the case here)

I tend to think of `IN` as being slow too - but you're right, that's not necessarily true, and using `EXISTS` would address that problem.

PR #226, opened 2025-03-26
fix(breadbox): data download page 25q2

naquib314:
This PR in the Data Downloads page:
1. Adds long reads under sequencing 
2. Moves ATACSeq under Sequencing


pipeline/data_page/get_all_data_availability.py
```diff
@@ -486,6 +553,12 @@ def main(
     )
     assert omics_summary.index.is_unique
 
+    # Long Reads
+    long_reads_summary = get_long_reads_summary(
+        gcloud_storage_client=gcloud_storage_client,
+        depmap_long_reads_gcloud_loc=depmap_long_reads_gcloud_loc,
+    )
```
alimourey, line 560: It looks like depmap_long_reads_gcloud_loc can be an empty list. Is that because long reads data isn't public yet?

pipeline/data_page/get_all_data_availability.conseq
```diff
@@ -4,25 +4,11 @@
...
-#        rppa_taiga_id=all {"type": "raw-rppa-matrix"},
-#        ms_ccle_taiga_id=all {"type": "proteomics-raw"},
-#        sanger_methylation_taiga_id=all {"type": "sanger_methylation_taiga_id"},
-#        methylation_ccle_taiga_id=all {"type": "biomarker-correctly-transposed"},
-#        ccle_mirna_taiga_id=all {"type": "ccle_mirna_taiga_id"},
-#        ataq_seq_taiga_id=all {"type": "ataq_seq_taiga_id"},
-#        olink_taiga_id=all {"type": "olink_taiga_id"},
-#        sanger_proteomics_taiga_id=all {"type": "sanger-proteomics"}
+        depmap_long_reads_gcloud_loc={"type": "depmap_long_reads_gcloud_loc"}, # Adding it separately since it's not a taiga id but a gcloud location
+        script=fileref("get_all_data_availability.py"),
```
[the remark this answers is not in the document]
pgm: Great observation!

I can clarify on the "It looks like depmap_long_reads_gcloud_loc might be non-public. Is that correct?" part.

Yes, that's correct there is no publicly released Long reads data. I likely forgot to communicate that when I described the requirements to Nayeem. 

Nayeem, it'll be important to make this gracefully handle the case where there is no such artifact.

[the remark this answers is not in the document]
naquib314: Sounds good.

pipeline/data_page/get_all_data_availability.py
```diff
@@ -486,6 +553,12 @@ def main(
     )
     assert omics_summary.index.is_unique
 
+    # Long Reads
+    long_reads_summary = get_long_reads_summary(
+        gcloud_storage_client=gcloud_storage_client,
+        depmap_long_reads_gcloud_loc=depmap_long_reads_gcloud_loc,
```
alimourey, line 559: For the public environment, I think this will fail. It looks like public would result in depmap_long_reads_gcloud_loc = []. And then inside this function there will a key error when it tries to find "bucket_name" etc. You could follow the same pattern used below for OncRef.
```

`    # should be None for public
    oncref_summary = None
    if len(depmap_oncref_taiga_id) > 0:
        oncref_summary = get_oncref_summary(
            tc=tc,
            depmap_oncref_taiga_id=f"{depmap_oncref_taiga_id[0]}/PRISMOncologyReferenceViabilityMatrix",
        )`
```

naquib314: Ah, good catch!

pipeline/data_page/get_all_data_availability.conseq
```diff
@@ -52,7 +36,16 @@ rule get_all_data_availability:
...
                 ('ms_ccle_taiga_id', 'proteomics-raw'), 
-                ('sanger_methylation_taiga_id', 'sanger_methylation_taiga_id'), ('methylation_ccle_taiga_id', 'biomarker-correctly-transposed'), ('ccle_mirna_taiga_id', 'ccle_mirna_taiga_id'), ('ataq_seq_taiga_id', 'ataq_seq_taiga_id'), ('olink_taiga_id', 'olink_taiga_id'), ('sanger_proteomics_taiga_id', 'sanger-proteomics'), ('depmap_paralogs_taiga_id', 'depmap_paralogs_taiga_id')]:
+                ('sanger_methylation_taiga_id', 'sanger_methylation_taiga_id'), 
+                ('methylation_ccle_taiga_id', 'biomarker-correctly-transposed'), 
+                ('ccle_mirna_taiga_id', 'ccle_mirna_taiga_id'), 
+                ('ataq_seq_taiga_id', 'ataq_seq_taiga_id'), 
+                ('olink_taiga_id', 'olink_taiga_id'), 
+                ('sanger_proteomics_taiga_id', 'sanger-proteomics'), 
+                ('depmap_paralogs_taiga_id', 'depmap_paralogs_taiga_id'),
+                ('sequencing_atacseq_broad_taiga_id', 'sequencing_atacseq_broad_taiga_id'),
```
alimourey, line 46: I think you accidentally added `('sequencing_atacseq_broad_taiga_id', 'sequencing_atacseq_broad_taiga_id')`

pipeline/data_page/get_all_data_availability.py
```diff
@@ -357,37 +358,103 @@ def get_paralogs_summary(tc, depmap_paralogs_taiga_id):
...
+    return long_reads_summary
+
+
+def get_taiga_id(possible_id, id_key="dataset_id"):
     return [] if len(possible_id) == 0 else [possible_id[0][id_key]]
 
 
+def get_gcloud_loc(possible_id):
+    if len(possible_id) == 0:
+        return []
```
pgm, line 417: This is a little unusual. I see that you're testing 
```
    if len(depmap_long_reads_gcloud_loc) > 0:
```
which is technically going to work but it's very werid that depmap_long_reads_gcloud_loc is either type of list *or* a dict, and if it's a list, it's always an empty list.

How about changing this to `None` instead? (and the corresponding test to check for `...is not None`)

As written, it looked like a mistake because it's uncommon to have mixed return types like that.

pipeline/data_page/get_all_data_availability.py
```diff
@@ -357,37 +358,103 @@ def get_paralogs_summary(tc, depmap_paralogs_taiga_id):
...
+
+    Returns:
+        pd.DataFrame: Indexed by ACHID or ModelID, with a column of True values that will be filterd against the Model table in future
+    """
+
+    print("getting long_reads_summary...")
+
+    bucket_name = depmap_long_reads_gcloud_loc["bucket_name"]
+    prefix = depmap_long_reads_gcloud_loc["prefix"]
+    file_names = depmap_long_reads_gcloud_loc["file_names"].replace(" ", "").split(",")
```
pgm, line 377: This will work, but you actually only need to read one file. The model IDs should be consistent across all files because we either have all types of data for the model or no data for the model. I don't think we should have a mix of different models in different files.

pgm (approved):

PR #227, opened 2025-04-02
Fix Genomic Associations tab

snwessel:
[Asana](https://app.asana.com/0/1165651979405609/1209815448333172/f)

Once this is merged, I'll cherry-pick it to the DMC branch and test that this fixes the issue in the staging environment before deploying to prod.

**Background on the problem:** 
* It seems like this tab is already working on the gene pages (for example [here](https://dmc.depmap.org/portal/gene/KRAS?tab=genomic_assoc)), so this bug seems to only affect the compound pages.  
* Genomic Associations hasn't worked locally for me for a long time, which I discovered was an [unrelated celery issue](https://app.asana.com/0/1202104943355905/1209860469728924/f) that doesn't seem to affect the deployed environments
* A little while ago, I did some refactoring of the compound pages ([here](https://github.com/broadinstitute/depmap-portal/pull/188)), and we'd been talking about removing/replacing Genomic Associations, so I hadn't been focused on testing how my changes affected it. Also it was broken for me locally, so I didn't have a great way of testing it. It looks like those refactoring changes did create this bug though. 
* The actual bug here was that our `format_celfie` function was being called with breadbox datasets, which it is not set up to handle. Instead of doing that, I am now passing in only the legacy datasets (which are indexed by compound experiment). 

I'm also including a couple other small fixes for potential issues I noticed while debugging. 

**This is now working for me locally:**
_Note: this tab is still showing compound experiments in the dataset-selection dropdown._ This is different from other parts of the compound page, which have been updated to show only compound-level information. I did look into updating this, but there would be quite a bit of refactoring required, which probably isn't worthwhile at the moment.
image


portal-backend/depmap/compute/views.py
```diff
@@ -172,9 +179,16 @@ def post(self):
...
-            query_series = query_series[~query_series.isna()]
+            if slice_query_is_from_breadbox:
+                query_series = data_access.get_slice_data(slice_query)
+            else: 
+                # In this specific case, it's important to avoid the data_access interface because 
+                # breadbox legacy-dataset aliases don't work for lookups by entity_id 
+                # (which is still the slice format used by celfie/genomic associations)
+                query_series = interactive_utils.get_row_of_values_from_slice_id(
+                    query_id
+                )
```
snwessel, line 190: I had initially thought the source of the issue was that custom analysis was being called with a dataset that was aliased in breadbox and a legacy "entity_id" which doesn't consistently map onto the IDs breadbox uses for features (for some datasets they do match, for others they don't). My concern was that our aliasing was automatically trying to look up values in breadbox, when some of the old slice IDs only work with legacy datasets. 

In reality, I don't think anything is being called this way, so it's probably not an issue, but I updated this function just in case. Now, if anyone calls custom analysis with a non-breadbox slice ID, it will avoid the possibility of aliasing.

depmap-compute/depmap_compute/slice.py
```diff
@@ -38,3 +38,26 @@ def decode_slice_id(slice_id) -> tuple[str, str, str]:
...
+    # Slice query identifier types use different (more descriptive) terminology
+    if slice_type == "label":
+        slice_query_identifier_type = "feature_label"
+    elif slice_type == "transpose_label":
+        # "transpose_label" identifiers are always depmap IDs
+        # In breadbox, this is stored as an ID, not a label.
+        slice_query_identifier_type = "sample_id"
+    elif slice_type == "entity_id":
+        # "entity_id" is only used in older parts of the codebase (not DE2 or ContextManager)
+        assert not dataset_id.startswith("breadbox/"), "Breadbox datasets do not support lookups by entity_id"
```
snwessel, line 56: Unrelated fix: I noticed that this function had gotten deleted somehow, even though it's imported by the `portal-backend`, which was only working because it was using an older version of this `depmap-compute `package that had the function. 

I also updated this function to ensure that breadbox datasets are never queried with the `"entity_id"` slice type (see my above comment for more background on what that's a potential issue). I _believe_ the only parts of our codebase that use the `entity_id` slice type at this point are:
* Genomic Associations 
* Constellation

depmap-compute/depmap_compute/slice.py
```diff
@@ -38,3 +38,26 @@ def decode_slice_id(slice_id) -> tuple[str, str, str]:
...
+
+def slice_id_to_slice_query(slice_id: str) -> SliceQuery:
+    """Take a legacy slice ID string and convert it to the newer slice query format."""
+    dataset_id, dimension_identifier, slice_type = decode_slice_id(slice_id)
+
+    # Slice query identifier types use different (more descriptive) terminology
+    if slice_type == "label":
+        slice_query_identifier_type = "feature_label"
+    elif slice_type == "transpose_label":
+        # "transpose_label" identifiers are always depmap IDs
```
rcreasi, line 51: It could also reference a screen ID like this:
https://cds.team/depmap/data_explorer_2/unique_values_or_range?slice_id=slice/breadbox%2F30e1e790-44a3-4af4-9b3b-0efee94c053e/SC-000001.KY01/transpose_label

rcreasi: That's not exactly relevant to this PR (I don't think you can access screen-level data through the Genomic Associations tool) but this function looks pretty generic :)

PR #229, opened 2025-04-04
Reposition Inferred Molecular Subtype in Context Manager

rcreasi:
Asana: https://app.asana.com/0/1165651979405609/1209804368349686/f

This modifies the UI to treat one-hot encoded matrices more gracefully. Instead of exposing the details of the encoding to the user, the columns are displayed as pseudo-categories they can choose from.

For now, only the Molecular Subtypes dataset behaves this way. I've done some experiments that suggest this could be a good approach for handling all of our one-hot encoded matrices.

### DIFF
--- a/frontend/packages/@depmap/data-explorer-2/src/components/DatasetMetadataSelector/BinaryColorsHelpTip.tsx
+++ b/frontend/packages/@depmap/data-explorer-2/src/components/DatasetMetadataSelector/BinaryColorsHelpTip.tsx
@@ -1,0 +1,48 @@
+import React from "react";
+import cx from "classnames";
+import { Tooltip } from "@depmap/common-components";
+import { pluralize } from "../../utils/misc";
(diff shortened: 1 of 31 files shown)


frontend/packages/@depmap/data-explorer-2/src/components/DatasetMetadataSelector/BinaryColorsHelpTip.tsx
```diff
@@ -0,0 +1,48 @@
...
+      id="BinaryColorsHelpTip"
+      content={
+        <div>
+          Because these {pluralize(sliceTypeLabel)} overlap, a unique color
+          cannot be assigned to each one.
+          <br />
+          <br />
+          Please select {indefiniteArticle} {sliceTypeLabel} of interest and it
+          will be colored as the in-group.
+        </div>
```
rcreasi, line 34: tooltip

PR #231, opened 2025-04-04
Remove old datasets

jessica-cheng:
Remove old CRISPR and RNAi datasets listed in [Asana task](https://app.asana.com/0/1188486168213297/1209788511255105) so that they don't appear in Data Explorer 2 or Custom Downloads

- Datasets that have a dependency enum defined are entirely removed from the codebase including the preprocessing pipeline
- Some of the datasets are nonstandard datasets which are defined elsewhere and they are removed in [this PR](https://github.com/broadinstitute/depmap-deploy/pull/70)

_NOTE: Currently we cannot remove RNAi_Ach and RNAi_Nov_DEM datasets since they are used in the downloads overview graph. Therefore, we filter those specific datasets out so they don't appear in DE2 and Custom Downloads_


portal-backend/depmap/interactive/interactive_utils/config_groups.py
```diff
@@ -34,8 +34,14 @@ def get_all_dataset_ids():
...
         - custom downloads
+        - de2
     """
-    all_dataset_ids = {dataset_id for dataset_id in __get_config().all_datasets}
+    # We do not want to show datasets for RNAi_Ach and RNAi_Nov_DEM in custom downloads and DE2. However, we cannot completely remove them from the codebase now since these dataset are referenced in the downloads overview graph so filter them out instead
+    all_dataset_ids = {
+        dataset_id
+        for dataset_id in __get_config().all_datasets
+        if dataset_id not in ["RNAi_Ach", "RNAi_Nov_DEM"]
+    }
```
alimourey, line 44: I have 2 suggestions for how to avoid this filtering!

1. For the data overview graph, the enum values are only used to determine the y-axis urls for "Achilles (Broad)" and "Drive (Novartis)". If we wanted to completely remove these 2 datasets from the enum, instead of filtering here, we could unlink the y-axis for those 2 datasets. I could do this, or point out to where to do this.

2. If we want to also completely remove the RNAi artifacts from preprocessing, we could also just switch get_all_data_availability.py to use hardcoded taiga ids for these 2 datasets. We already do this for the RNAi marcotte dataset.

This assumes we definitely want to keep these rows in the data overview graph. I think I could pretty quickly remove these rows from the graph entirely.

jessica-cheng: @pgm My understanding is that removing dataset mentions from DE2 and custom downloads effectively translates to removing them from the portal entirely (except for options from all historical downloads). The data overview graph seems to communicate the current release's available datasets within the portal so I think it makes sense to remove from the graph but I'll leave this decision of removing them from the data overview graph up to you

jessica-cheng: Decision: We want to keep references to those RNAi datasets within the data overview graph.

PR #233, opened 2025-04-07
perf(breadbox): Etag setup

snwessel:
[Asana](https://app.asana.com/0/1165651979405609/1209782696163971/f)

You can try this out yourself here: https://dev.cds.team/depmap-test-perf/breadbox/elara/

A full page load with pre-cached identifiers reloads in ~2 seconds now, even for a large gene-indexed dataset ([here](https://dev.cds.team/depmap-test-perf/breadbox/elara/?p=eJxVjsFqwzAQRP9lzzHEsWKnhhza_kGvJYiNd20EtiwkkSoY_XtWgbT0Imk0b3ZnAzevUce7Y-iB2AYT77om2IGxxOnPcQs6vazEs3hkloKuNkC_QSoHJhNetMcfHWYzsKA4TZ4njAKLMxofovwOq42cYgn-znqlJ7YlaHEp6uPr870Wycn5gp_P0H9vcENfUHNjqw1B3kHbHeAitzilVZYnYcTAsQA9qI6a9m3fVHV3GCt1bKk6ccPVqUGlaLgSqlH2PHv_q5JzfgAtRWEc)) .

**In comparison:**
* When we started making these performance improvements, this same page load in Elara was taking about 6 seconds
* In non-elara DE2 ([here](https://dev.cds.team/depmap-test-perf/data_explorer_2/?p=eJxVjtFqwzAMRf_Fzw0jbbaaQB_W_cFexwiKJQdDYhtbbC7B_z55tIy9CF3dc9HdVVwDT3yLpEaF5LPj29SjOijnkcqfEzeI0xaQVvHQbQ0NPqtxV6UNKC4_6ATfU16dIUFhWRItwAKLY13KLFcTPFPhFryvj-xCvsU8bE1d399ee5FUYmrw5aLGj119gShFnlvbFWYpVQ93-LPKisCQiSeHws2JAOdQnrTWaOzz0Bl8sd1wOg6dxvPQWTr2pj_Zs0Yrz36r_-tTa_0BpehjFQ)), this same page load is still taking about 11 seconds
So in total, this is MUCH improved!

But it does also add to the complexity of this endpoint.

### DIFF
--- a/breadbox/breadbox/api/dimension_types.py
+++ b/breadbox/breadbox/api/dimension_types.py
@@ -652,17 +652,21 @@ def get_dimension_type_identifiers(
             data_type=data_type,
         )
         filtered_dataset_ids = [dataset.id for dataset in filtered_datasets]
-    etag = hash_id_list([name] + (sorted(filtered_dataset_ids) if filtered_dataset_ids else []))
+
+        if show_only_dimensions_in_datasets:
+            # Remove the metadata dataset if it's in our list of datasets 
+            filtered_dataset_ids = [dataset_id for dataset_id in filtered_dataset_ids if dataset_id != dim_type.dataset_id]
+    
+    # Create an ETag based on the dimension type and the filtered dataset IDs
+    etag = hash_id_list(
+        [dim_type.name] +
+        [dim_type.dataset_id] if dim_type.dataset_id else [] + 
+        (sorted(filtered_dataset_ids) if filtered_dataset_ids else [])
+    )
 
     def _get_response_content() -> list[DimensionIdentifiers]:
-        if filtered_dataset_ids is None:
-            dataset_ids_without_metadata = None
-        else:
-            # Remove the metadata dataset from our list of datasets 
-            dataset_ids_without_metadata = [dataset_id for dataset_id in filtered_dataset_ids if dataset_id != dim_type.dataset_id]
-            
         dimension_ids_and_labels = metadata_service.get_dimension_type_identifiers(
-            db, dim_type, dataset_ids_without_metadata, limit=limit,
+            db, dim_type, filtered_dataset_ids, limit=limit,
         )
         return [
             DimensionIdentifiers(id=id, label=label)
--- a/breadbox/tests/api/test_types.py
+++ b/breadbox/tests/api/test_types.py
@@ -475,7 +475,10 @@ def test_get_dimension_type_dimension_identifiers(
     )
     assert res.json() == []
 
-    # Test case if data type is metadata and show only dimensions in datasets is True when there is a dataset of data type metadata with requested dimension type but is not used as a dimension types's metadata. The result should still filter out identifiers that are not found in the dimension type's metadata
+    # Test case if data type is metadata and show only dimensions in datasets is True when 
+    # there is a dataset of data type metadata with requested dimension type but is not used 
+    # as a dimension types's metadata. 
+    # The result should still filter out identifiers that are not found in the dimension type's metadata.
     matrix_values = factories.matrix_csv_data_file_with_values(
         feature_ids=["feature-1", "feature-2", "feature-3"],
         sample_ids=["sample-1", "sample-3"],
@@ -496,7 +499,11 @@ def test_get_dimension_type_dimension_identifiers(
         {"id": "sample-1", "label": "Sample 1"},
     ]
 
-    # Test case if data type is metadata and show only dimensions in datasets is False when there is a dataset of data type metadata with requested dimension type but is not used as a dimension types's metadata. This should return dimension identifiers from both datasets but should still filter out identifiers that are not found in the dimension type's metadata
+    # Test case if data type is metadata and show only dimensions in datasets is False when 
+    # there is a dataset of data type metadata with requested dimension type but is not used 
+    # as a dimension types's metadata. 
+    # This should return dimension identifiers from both datasets but should still filter out 
+    # identifiers that are not found in the dimension type's metadata.
     res = client.get(
         f"types/dimensions/{dim_type_fields['name']}/identifiers?data_type=metadata",
         headers=admin_headers,


breadbox/breadbox/api/dimension_types.py
```diff
@@ -622,24 +624,56 @@ def list_dimension_types_endpoint(db: SessionWithUser = Depends(get_db_with_user
...
-    dimension_ids_and_labels = metadata_service.get_dimension_type_identifiers(
-        db, dim_type, data_type, show_only_dimensions_in_datasets, limit=limit,
-    )
-
-    return [
-        DimensionIdentifiers(id=id, label=label)
-        for id, label in dimension_ids_and_labels.items()
-    ]
+    # If the client already has a cached version of the data, exit early
+    if if_none_match and if_none_match[0] == etag:
```
pgm, line 659: what do you think about something like:

```
def get_etag_handler(etag, 
  generate_response, 
  if_none_match: Optional[List[str]] = Header(None)):

  if if_none_match and if_none_match[0] == etag:
    return get_not_modified_response(etag)

  result = generate_response()
  return get_response_with_etag(content=result, etag=etag)
  
def get_dimension_type_identifiers(
    request: Request,
    name: str,
    data_type: Annotated[Union[str, None], Query()] = None,
    show_only_dimensions_in_datasets: Annotated[bool, Query()] = False,
    limit: Annotated[Union[int, None], Query()] = None,
    db: SessionWithUser = Depends(get_db_with_user),
    etag_handler = Depends(get_etag_handler)):
...
   etag = hash_id_list([name] + (sorted(filtered_dataset_ids) if filtered_dataset_ids else []))

   def _get_response():
      if filtered_dataset_ids is None:
            dataset_ids_without_metadata = None
        else:
      ...
      return result

   return etag_handler(etag, _get_response):
```

I'm trying to lift as much of the handling of etag out of the endpoint. The version above gets us to the least amount of boilerplate and a pattern that we could copy for other endpoints.

breadbox/breadbox/api/dimension_types.py
```diff
@@ -622,24 +624,56 @@ def list_dimension_types_endpoint(db: SessionWithUser = Depends(get_db_with_user
...
+        return get_not_modified_response(etag)
+    else:
+        if filtered_dataset_ids is None:
+            dataset_ids_without_metadata = None
+        else:
+            # Remove the metadata dataset from our list of datasets 
+            dataset_ids_without_metadata = [dataset_id for dataset_id in filtered_dataset_ids if dataset_id != dim_type.dataset_id]
```
pgm, line 666: This could be included as part of the list comprehension above correct? (I'm not sure I fully understand what this is for)

breadbox/breadbox/api/dimension_types.py
snwessel: I actually had an earlier version of this branch working with a very similar setup to what you suggesting (you can see it [here](https://github.com/broadinstitute/depmap-portal/blob/6e9f024b3893b04470fbc642d86963543a948099/breadbox/breadbox/api/dimension_types.py#L660)), and I'd be happy to move back to that if you think that's easier to reuse for other endpoints. The reason I'd moved away from that approach was that it involved a more boilerplate than you have above. 

I believe the implementation you have above creates some issues with variable scoping when the function `_get_response` is actually called.  For example, the internals of `_get_response` wouldn't have access to variables like `db` or `filtered_dataset_ids` unless they're explicitly passed in by defining `_get_response` with a bunch of parameters, and then using a partial like:

```
callback = partial(_get_response, db=db, dim_type=dim_type, filtered_dataset_ids=filtered_dataset_ids)
```

Alternatively, you could just define the variables in `_get_response` as:
```
nonlocal db
nonlocal dim_type
...
```

Either way is a fine solution, it's just a bit more boilerplate and makes things a bit more difficult to follow. If you think that approach is easier to read and/or reuse, I'd be happy to switch back to that - I don't really have a strong preference one way or another.

pgm: This is very surprising. Python will create a `_get_response` closure for that function. `nonlocal` is only needed if you want to support mutation in both inner and outer scope (which isn't the case here)

breadbox/breadbox/api/dimension_types.py
```diff
@@ -622,24 +624,56 @@ def list_dimension_types_endpoint(db: SessionWithUser = Depends(get_db_with_user
...
     name: str,
     data_type: Annotated[Union[str, None], Query()] = None,
     show_only_dimensions_in_datasets: Annotated[bool, Query()] = False,
     limit: Annotated[Union[int, None], Query()] = None,
     db: SessionWithUser = Depends(get_db_with_user),
+    if_none_match: Optional[List[str]] = Header(None), # etag from the client's cache
 ):
+    """
+    For the given dimension type and filters, get all given IDs and labels.
+    If the `data_type` is given and/or `show_only_dimensions_in_datasets` is True, the dimension identifiers that are returned will only be those that are used within a dataset.
```
jessica-cheng, line 637: I noticed this comment is copied from my previous implementation and realized that it is somewhat misleading. It _looks like_  the dimension identifiers that are returned will only be those that are used within a dataset when `data_type` is given and `show_only_dimensions_in_datasets = False` because usually there is an implicit filtering that is happening since we usually want to filter by a `data_type` that is not `metadata`. However, that is not always the case such as when `data_type=metadata` which might become an issue when we start to create datasets where the data type can be `metadata` even though it is not used as a dimension type's metadata.

I noticed that it seems like you are trying to refactor my original implementation here and while I do like some of the refactoring done here, I noticed some of the assumptions you made may be incorrect. See [my PR for new test cases I added](https://github.com/broadinstitute/depmap-portal/pull/234). These would fail with your implementation but pass with my original implementation.

jessica-cheng (commented):
I think the overall logic for etag support seems fine but there are some assumptions done in your refactoring which may be incorrect. Let me know if you have any questions!

PR #235, opened 2025-04-08
Dimension type identifiers test fix

jessica-cheng:
Fix test for dimension type identifier endpoint and add more test cases

**NOTE: ** _These new test cases will fail with your refactoring changes but should pass for my original implementation_


pgm:
There are conflicts and I'm 100% sure I know what this test trying to achieve. I'm going to close it until we find a need for this.

PR #238, opened 2025-04-11
Switch portal pipeline to retrieve crispr common essential from taiga instead of calculating 

naquib314:


pipeline/scripts/tda/CE_percentile_rank_analysis.py
```diff
@@ -1,41 +1,79 @@
...
+    import sys
+
+    data_label = sys.argv[1]
+    crispr_inferred_common_essentials_file = sys.argv[2]
+    non_crispr_input_data_file = sys.argv[3]
+    out_file = sys.argv[4]
+
+    # Crispr is retreived from taiga
+    # Non-crispr is calculated from the data file
+    if data_label == "Chronos_Combined":
```
pgm, line 74: I personally would have gone the route of having two different conseq rules: one for RNAi (which would be the original rule, running the original script) and one for Chronos_Combined which reformats the file. 

It feels clearer to me. For example, if I understand how this will work, when run on RNAi, the rule is going to have a dependency on crispr_inferred_common_essentials_file even when that file is ignored.

Anyway, I don't think it's worth changing anything because I think this is a change that is largely isolated from anything that'll we'll be touching anytime soon.

Just sharing a different approach.

pipeline/tda.conseq
```diff
@@ -16,14 +16,39 @@ rule format_gene_dep_csv:
         df.to_csv("deps.csv",index_label="Row.name")
     """
 
+# This rule only exists to get the CRISPRInferredCommonEssentials dataset from taiga and save it to a file
+# so that it can be used by the common_essentials rule later. It is being done this way because the common_essentials
+# rule is being run inside a docker container, which doesn't have taigapy installed and it seems to be the 
+# least painful way to get around that other than not using dsub for common_essentials rule.
```
pgm, line 22: Here's another reason why having a different rule for combined_chronos would probably would have worked better. One could have had the new rule not rely on that docker container.

pgm (approved):

pgm:
Already merged in.

PR #240, opened 2025-04-11
fix(breadbox): Pin python versions, simplify docker setup

snwessel:
[Asana](https://app.asana.com/1/9513920295503/project/1165651979405609/task/1209914649715514?focus=true)

**I started looking into this because:**
1. I noticed that for a while, the breadbox github action has shown fewer pyright errors than I see locally
2. I wanted to bump the pyright version because it was creating difficult to-debug local errors for me (which is fixed by newer versions)

**Narrowing down the problem:**
1. I initially thought the inconsistency was coming from a difference in python versions. We've talked about pinning the python version anyway, and Devin already has a good workflow for this, so I copied his workflow of using pyenv. However, this did _not_ fix the pyright issues. 
2. I noticed that the breadbox `Dockerfile` does a bunch of seemingly unnecessary steps in copying files and setting up poetry. This file hasn't been meaningfully changed in ~2 years, so I'm not sure why things were set up this way. After simplifying things to use a more standard approach, the pyright discrepancy went away.


breadbox/Dockerfile
```diff
@@ -1,45 +1,17 @@
-FROM us.gcr.io/broad-achilles/depmap-breadbox-base:v2
+FROM us.gcr.io/broad-achilles/depmap-breadbox-base:v2
```
snwessel, line 1: I'm curious how/where this base image is defined. It's probably setting the python version for our docker build. But I haven't been able to find the Dockerfile it's built from. 

It is possible to set the docker python version using pyenv ([example here](https://gist.github.com/jprjr/7667947)), but it's not the most elegant thing. From my understanding, the standard/simpler approach is to set the python version in your base image ([like this](https://github.com/broadinstitute/snp_str_qc/blob/develop/Dockerfile)) and then use pyenv to manage the python version for local development.

breadbox/.python-version
```diff
@@ -0,0 +1 @@
+3.9.15
```
snwessel, line 1: For now, I've just set this to match the version that the breadbox docker image already uses.

breadbox/pyright-ratchet-errors.txt
```diff
@@ -1,21 +1,29 @@
...
+dataset.py: error: Argument of type "type[DatasetSample]" cannot be assigned to parameter "axis" of type "DatasetFeature | DatasetSample" in function "_get_indexes_by_given_id"
+dataset.py: error: Type "Series | DataFrame* | DataFrame" is not assignable to return type "DataFrame"
+dataset_tasks.py: error: Type "None" is not assignable to declared type "List[Unknown]"
+dataset_uploads_tasks.py: error: Argument of type "Series | DataFrame* | DataFrame" cannot be assigned to parameter "given_ids" of type "Sequence[str]" in function "_get_dimension_labels_and_warnings"
+download_tasks.py: error: Argument of type "ArrayLike" cannot be assigned to parameter "column_labels" of type "List[str] | None" in function "_handle_df_nas"
+download_tasks.py: error: Cannot access attribute "merge" for class "NDArray[Unknown]"
+download_tasks.py: error: Cannot access attribute "merge" for class "ndarray[Any, Unknown]"
+download_tasks.py: error: Cannot access attribute "merge" for class "ndarray[Any, _dtype]"
+env.py: error: Argument of type "str | PathLike[str] | None" cannot be assigned to parameter "fname" of type "StrOrBytesPath | IO[str] | RawConfigParser" in function "fileConfig"
```
snwessel, line 22: Some of the differences here are just wording changes in the error descriptions. But the newer version of pyright does also generate a few new errors.

breadbox/Dockerfile
pgm: This was built using the `Dockerfile` in `depmap/depmap-base`. 

It was originally created to speed up building the final image, by starting with a docker image that had things that _rarely_ changed pre-installed. In particular *R and it's dependencies used to take > 30 minutes to install*. 

However, now that we no longer depend on R, it's entirely possible that we could stop using this. I _suspect_ the remaining dependencies are fairly quick to install.

pgm (approved):
Looks good

breadbox/.python-version
jessica-cheng: Just out of curiosity, why do you define the python version here instead of pinning it in the `pyproject.toml` file?

snwessel: Originally, I'd been planning to pin it in the pyproject.toml file, but the more I looked into it, it seemed like `pyenv` was the more generally accepted/recommended approach. Here's what the [official poetry documentation](https://python-poetry.org/docs/managing-environments/) says about python versions:

> By default, Poetry will try to use the Python version used during Poetry’s installation to create the virtual environment for the current project.
However, for various reasons, this Python version might not be compatible with the python range supported by the project. In this case, Poetry will try to find one that is and use it. If it’s unable to do so then you will be prompted to activate one explicitly, see [Switching environments](https://python-poetry.org/docs/managing-environments/#switching-between-environments).

My takeaway from that is that poetry is able to use a specific python version if that version happens to already be installed on your computer. If it's not already installed, you'll need to use `pyenv` to install the specific version anyway. I'm trying to be a bit more proactive by using `pyenv` in our `install_prereqs` script to make sure everyone has the correct version and is using it in their poetry environment.

More generally, my impression from reading the documentation is that poetry isn't really intended to specify a specific version of python (ex. it's not really intended for specifying 3.9.15 vs. just python 3 or python 3.9). Most of the examples I see suggest using `pyenv` for pinning a more specific version.

jessica-cheng (approved):
LGTM

PR #247, opened 2025-04-22
Remove old private dataset frontend & loader

snwessel:
[Asana](https://app.asana.com/1/9513920295503/project/1165651979405609/task/1209992967754051?focus=true)

First pass at removing a bunch of private dataset functionality, trying not to impact the transient datasets that are still uploaded and stored here. 

**The things I'm removing here are:**
* All of the frontend code for the old UI for uploading private datasets 
* The few API endpionts called by that frontend:
    * `GET /access_groups`
    * `GET /interactive/from-csv-url`
    * `GET /private/<bucket_name>/<path:filename>`
    * `POST /private` (for uploading private datasets from the primary upload UI) 
    * `GET private_dataset/upload_status/<task_id>`
    * `DELETE private_dataset/delete`
* The `synchronize_private_datasets` step that was run as a part of the database loader
* Lower-level utils that were no longer being used (ex. `data_access.get_private_datasets`) 
* Lots of old tests


pgm:
I think `/interactive/from-csv-url` might be important. Let me check.

snwessel:
> I believe we cannot drop /interactive/from-csv-url. It's used for linking to the DE2 from the PRISM portal.
As a result, you won't find any references to this inside the depmap-portal repo, only in the PRISM portal. (At the very least, we probably should have added a comment to reflect that because it's going to look unused.)

Good to know, thank you! I'll add it back in and add a comment

pgm:
I think we'll need to add back that endpoint, but excited to see code being deleted!

image

PR #249, opened 2025-04-23
Part 2 of removing old private dataset system

snwessel:
[Asana](https://app.asana.com/1/9513920295503/project/1165651979405609/task/1209992967754051?focus=true)

**Now that the private datasets have been migrated to breadbox and are no longer being loaded into the legacy DB, it should be safe to remove these old access controls.**

I'm also removing a few other things that had been supporting the legacy private datasets:
* private dataset lookup steps from our legacy `get_datasets`/`get_dataset` methods
* the `PRIVATE_FILE_BUCKETS` setting, which isn't being used anymore
* the `add_custom_taiga_dataset` endpoint, which had previously been used by data explorer 1, but is no longer being used
* our "canary_custom_dataset", which has historically been added on DB load

**After these changes, I tested that the following still work:**
1. The `recreate_dev_db` command runs
2. All tests are passing
3. The data explorer frontend + APIs are working as expected (generally, just from clicking around) 
4. I confirmed that a transient dataset uploaded through the DE2 User Upload modal is able to be plotted on DE2 and is not visible in other places (i.e. it's not appearing in custom downloads or as an option in the DE2 dataset selection) 

Note: There will need to be a "part 3" of this as well - once we have all the transient datasets being uploaded to breadbox. I'll add that to Asana separately.


portal-backend/depmap/cas/util.py
```diff
@@ -12,7 +12,7 @@ def get_storage_client() -> storage.Client:
     return client
 
 
-def get_private_datasets_bucket(client: storage.Client = None) -> storage.Bucket:
+def get_cas_bucket(client: storage.Client = None) -> storage.Bucket:
     if client is None:
         client = get_storage_client()
     bucket_name = current_app.config["CAS_BUCKET"]
```
snwessel, line 18: I'm not sure why this was named this before - it doesn't seem to be doing anything related to private datasets.

portal-backend/depmap/settings/download_settings.py
```diff
@@ -17,6 +17,7 @@ def get_download_list() -> List[DownloadRelease]:
     # now these releases may include things that the current user is not allowed to see, so filter the list
     owner_ids = set(get_visible_owner_id_configs().keys())
 
+    # TODO: not sure what an access controlled release would be? I assume I don't need to worry about this?
     releases: List[DownloadRelease] = [
         release for release in download_releases if release.owner_id in owner_ids
```
snwessel, line 22: @pgm this looks like some sort of release-level access control which I don't believe we're using - but I'm also not that familiar with how the releases are organized in the portal. Do you know of any reason that this filtering should be kept?

pgm (approved):
LGTM

PR #256, opened 2025-04-29
Fix delete dataset button state

jessica-cheng:
- Fix error where deleting dataset does not clear state such that delete button is disabled ([see error log here](https://app.asana.com/1/9513920295503/project/1188486168213297/task/1210082114924877))
- Show delete error message for 3 seconds (it's a bit confusing the error message stays around when user does some other action)
- remove unnecessary state

### DIFF
--- a/frontend/packages/@depmap/dataset-manager/src/components/index.tsx
+++ b/frontend/packages/@depmap/dataset-manager/src/components/index.tsx
@@ -182,7 +182,7 @@ export default function Datasets() {
     // Only show dim type delete error message for 3 seconds
     if (dimTypeDeleteError) {
       setTimeout(() => {
-        setDatasetDeleteError(null);
+        setDimTypeDeleteError(null);
       }, 3000);
     }
   }, [dimTypeDeleteError]);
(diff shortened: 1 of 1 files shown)


frontend/packages/@depmap/dataset-manager/src/components/index.tsx
```diff
@@ -410,26 +426,26 @@ export default function Datasets() {
...
-          (dataset) => !datasetIdsSet.has(dataset.id)
+          (dataset) => !selectedDatasetIds.has(dataset.id)
         );
         setDatasets(datasetsRemaining);
         const dimTypeDatasetsNum = dimensionTypeDatasetCount(datasetsRemaining);
         const updatedDimensionTypeCounts = dimensionTypes.map((dt) => {
           return { ...dt, datasetsCount: dimTypeDatasetsNum[dt.name] };
         });
         setDimensionTypes(updatedDimensionTypeCounts);
+        setSelectedDatasetIds(new Set());
```
snwessel, line 446: Just from reading through this, it looks like this waits to disable the button until after all of the selected datasets have been deleted, is that right? 

I'm not sure how long the deletions usually take, but I feel like I remember it being a bit slow in the past. If it takes a couple of seconds to delete several datasets, it might be worth disabling the button sooner so that the user has some indication that their request is being processed (and doesn't try to immediately click again). For example, minigumbo has something similar which wasn't too difficult to implement (see [here](https://github.com/broadinstitute/minigumbo/blob/7691ba7253004517b0590edf8beee3d881b1b1c7/frontend/src/components/table/ServerBackedTable.tsx#L184)):

https://github.com/user-attachments/assets/374d0829-669b-442f-9801-63ca13da13dc

snwessel (approved):
Looks good, thanks for taking a look at this! And it's always nice when you're able to simplify things a bit in the process of improving the UX. 

I just had one small comment/question, but otherwise this looks good 👍

PR #257, opened 2025-04-30
Fix current release data page subtypes

alimourey:
Task: https://app.asana.com/1/9513920295503/project/1165651979405609/task/1208838500242933
Support for parsing downloads yaml with a subtypes mapping at the release level:

subtypes:
&nbsp;&nbsp;&nbsp;&nbsp;-code: this_is_the_code1
&nbsp;&nbsp;&nbsp;&nbsp;label:   This Is The Code1
&nbsp;&nbsp;&nbsp;&nbsp;-code: this_is_the_code2
&nbsp;&nbsp;&nbsp;&nbsp;label:   This Is The Code2

Individual files should also have a sub_type with a value that is one of the codes from the subtypes mapping:

sub_type: this_is_the_code

depmap-deploy: https://github.com/broadinstitute/depmap-deploy/tree/new-download-yaml-with-subtype-section
- Changes to the old yaml to support the new format. Not in a PR because the subtypes mapping and sub_type field only matters for the current release. I do have a script saved on my machine that will update existing yaml and add the subtype mapping, so I can use that on the 25q2 releases if we need to.

depmap-release-readmes PR: https://github.com/broadinstitute/depmap-release-readmes/pull/21


portal-backend/depmap/settings/parse_downloads.py
```diff
@@ -98,8 +112,8 @@ def make_file(file: Dict[str, Any]) -> DownloadFile:
     # Required for DownloadFile
     name = file.get("name", "")
     type = FileType(file.get("type", ""))
-    sub_type_val = file.get("sub_type", None)
-    sub_type = FileSubType(sub_type_val) if sub_type_val != None else sub_type_val
+
+    sub_type = get_file_sub_type(file.get("sub_type", None))
```
pgm, line 116: Looking at this implementation I notice there'll be a fair amount of redundancy (every code will need to have also have label defined)

Concretely, if we have a list of files it'll look like:
```
...
- name: ScreenSequenceMap.csv
  description: "..."
  sub_type: 
    code: model_conditions_mapping
    label: Model Conditions Mapping  
  type: genetic_dependency
- name: OmicsStructuralVariants.csv
  sub_type: 
    code: model_conditions_mapping
    label: Model Conditions Mapping  
  type: omics
  description: "..."
...
```

Looking back on what was written up in Asana, I see my proposal was very brief and so I think it wasn't clear what I was describing.

However, what I was proposing is that we add a property at the top of the release structure which lists out the labels for each code:

```
subtypes:
  - code: model_conditions_mapping
    label: Model Conditions Mapping
  - code: crispr
    label: CRISPR
...
- name: ScreenSequenceMap.csv
  description: "..."
  sub_type: model_conditions_mapping
  type: genetic_dependency
- name: OmicsStructuralVariants.csv
  sub_type: 
  type: omics
  description: "..."
...
```

This has a few advantages:
1. you can use the order of the `subtypes` property to determine what order to render the tabs. This would simplify some of the logic in `frontend/packages/portal-frontend/src/dataPage/components/ReleaseTabs.tsx` because it no longer would need a hardcoded list of ordering and have to handle other categories specially.
2. it minimizes the change to the YAML file structure. We only need to the `subtypes` property at the top. This point is moot because I see you have a PR containing changes to the release files, but I wanted to point it out that benefit because that was my thinking at the time.

Now, the question is: Is this worth changing? Your changes do achieve the big picture goal of allowing us to define new subtypes on demand, so is it worth changing?

I personally think it is worthwhile. My thinking is it's going be harder for folks to fill this out when they are adding/removing/recategorizing files in the potato editor. I also think it is useful to remove as many assumptions about what the sub_type values are. You've removed most of them, but I think it'll be good to also remove the hardcoded list that's being used for ordering. Then `read_me` will be the only assumption I can see, and that seems fair to have hardcoded.

alimourey: Makes sense. I just pushed the changes to redo this to fit your update proposal.

pgm (approved):
LGTM

PR #260, opened 2025-05-01
Update cell line description tile

alimourey:
These changes update the cell line page description tile to include every level of the new SubtypeTree. Each level links out to the appropriate page of Context Explorer V2. 

I kept the annotations from the Model file. Barbara and Alison had asked me to keep these in the Context Explorer data availability table, so I figured it's still useful to have here, but @pgm let me know if I should just remove them instead.

I had to rearrange the columns to fit everything cleanly. Now the tile looks like this (see action screenshot in Asana: https://app.asana.com/1/9513920295503/project/1165651979405609/task/1209731092974268):

### COLUMN 1:

**Annotations**

Oncotree Subtype and Code
oncotree_subtype

Oncotree Primary Disease
oncotree_primary_disease

Oncotree Lineage
oncotree_lineage

**Lineage Contexts**

Level 0
level_0

Level 1
level_1

### COLUMN 2:
**Molecular Subtypes**

Level 0
level_0

Level 1
level_1

**Should a header go over these? What should it say?**
Primary/Metastasis
prim_metastasis

Collection Site
collection_site

Source
source

Catalog Number
catalog_number

### DIFF
--- a/portal-backend/depmap/cell_line/views.py
+++ b/portal-backend/depmap/cell_line/views.py
@@ -1,5 +1,8 @@
+from dataclasses import dataclass
+import dataclasses
 import json
 from typing import Any, Dict, List
+from depmap.context.models_new import SubtypeContext, TreeType
 from depmap.enums import DataTypeEnum, TabularEnum
 from depmap.oncokb_version.models import OncokbDatasetVersionDate
 from flask import (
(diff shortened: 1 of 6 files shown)


portal-backend/depmap/cell_line/views.py
```diff
@@ -123,6 +126,35 @@ def get_related_models(patient_id: str, model_id: str) -> List[Dict[str, str]]:
...
+                subtype_code=row["subtype_code"],
+                level=row["node_level"],
+                context_explorer_url=url_for(
+                    "context_explorer.view_context_explorer",
+                    context=row["subtype_code"],
+                ),
+            )
+        )
+
+    tree_info = list(level_info.apply(create_subtype_level_dict, axis=1))
```
pgm, line 153: Not suggesting that this needs to change at all but wanted to share my thoughts on this because it looked a little unusual to me.

I thought it was curious that `SubtypeTreeInfo` is constructed only to by converted into a dict seemed like extra work, but then I guessed it's maybe to enforce typing? (ie: make sure that all the right property names are being used)

The other thing that caught my eye was the use of `list(level_info.apply(...))`. 

Pandas has a bunch of idioms that work well with dataframes, and I tend to write python as either using pandas idioms (encourages vectorized operations) or using python idioms (encourages sequential iterations and loops) and using `df.to_dict()` when I want to switch from pandas -> python.

Instead of:
```
tree_info = list(level_info.apply(create_subtype_level_dict, axis=1))
```
I'd write:
```
tree_info = [ create_subtype_level_dict(x) for x in level_info.to_dict("records") ]
```

I think it's a bias I have to prefer python patterns over pandas patterns, but I think at the end of the day, I can't see a reason to really advocate one over the other. I guess I'll say this one is a matter of personal preference.

snwessel (approved):
Looks good to me!

PR #262, opened 2025-05-02
Subtype Tree: Genetic subtypes whitelist as one input

acameron13:
Changing the conseq inputs to the subtype tree to provide the genetic subtypes whitelist as one static input, rather than using the latest version from taiga. This helps reproducibility and also will help us ensure that BLL -> BALL and TLL -> TALL as requested by PedDep.


pgm (approved):
Looks good!

PR #265, opened 2025-05-07
Fix enrichment box axis min and max

alimourey:
These box plots show up on both the Gene/Compound overview pages and in Context Explorer.

Previously, the plots were only using significant boxes to calculate their axis range.

These changes include the "other" (aka insignificant) data in the axis range calculations.

I also deleted a random console.log that had been left behind unnoticed.


alimourey:
@pgm This PR fixes the axis range issue Bella reported in the depmap-data-release Slack channel. I wasn't sure if we want this in internal for the release, or if you want to wait on adding this change. Let me know what you think!

pgm:
This sounds low risk so please merge it into the internal branch. Thank you

PR #268, opened 2025-05-08
Wide table selection prop

jessica-cheng:
Making this change to the WideTable component to support my Correlation Analysis changes. I believe that this change to how the prop `selectedLabels` is used in `ReactTableV7` would ultimately affect all components that use the `WideTable` prop `selectedTableLabels` since its value is passed into `selectedLabels`.   My understanding is that `selectedTableLabels` is used for a more fine-grained control over what gets selected in the table. I'm not too familiar with how this prop works/interacts with other components that use it. So far it appears it is mainly used in conjunction with the `WideTable` prop `singleSelectionMode` EXCEPT for `CellignerPage` which I am unsure how my change would affect its current selection behavior so I would definitely like a second set of eyes to make sure it works as expected.

These are the other components that use the `selectedTableLabels` prop (which I don't think my change should affect its current selection behavior):
- CompoundDashboardTable.tsx
- ContextAnalysisTable.tsx

### DIFF
--- a/frontend/packages/@depmap/wide-table/src/ReactTableV7.tsx
+++ b/frontend/packages/@depmap/wide-table/src/ReactTableV7.tsx
@@ -166,7 +166,7 @@ const ReactTableV7 = React.forwardRef(
           let newSelections: Set<any>;
           if (singleSelectionMode) {
             // We should always be getting the first selected label here since we expect selectedLabels size to be 1 in single selection mode
-            const label = [selectedLabels][0];
+            const label = [...selectedLabels][0];
             if (!prevSelections.has(label)) {
               newSelections = new Set([label]);
             } else {
--- a/frontend/packages/@depmap/wide-table/src/ReactTableV7.tsx
+++ b/frontend/packages/@depmap/wide-table/src/ReactTableV7.tsx
@@ -164,17 +164,23 @@ const ReactTableV7 = React.forwardRef(
       if (selectedLabels && selectedLabels.size > 0) {
         setSelections((prevSelections) => {
           const newSelections = new Set(prevSelections);
-          if (!newSelections.has([...selectedLabels][0])) {
+          if (
+            singleSelectionMode &&
+            !newSelections.has([...selectedLabels][0])
+          ) {
             newSelections.clear();
             newSelections.add([...selectedLabels][0]);
+          } else {
+            selectedLabels.forEach((selectedLabel) =>
+              newSelections.add(selectedLabel)
+            );
           }
-
           return newSelections;
         });
       } else {
         setSelections(() => new Set());
       }
-    }, [selectedLabels, idProp]);
+    }, [selectedLabels, idProp, singleSelectionMode]);
 
     const headersRef = useRef<HTMLDivElement | null>(null);
     const bodyRef = useRef<HTMLDivElement | null>(null);
--- a/frontend/packages/@depmap/wide-table/src/ReactTableV7.tsx
+++ b/frontend/packages/@depmap/wide-table/src/ReactTableV7.tsx
@@ -166,7 +166,7 @@ const ReactTableV7 = React.forwardRef(
           let newSelections: Set<any>;
           if (singleSelectionMode) {
             // We should always be getting the first selected label here since we expect selectedLabels size to be 1 in single selection mode
-            const label = [...selectedLabels][0];
+            const label = [selectedLabels][0];
             if (!prevSelections.has(label)) {
               newSelections = new Set([label]);
             } else {
--- a/frontend/packages/@depmap/wide-table/src/ReactTableV7.tsx
+++ b/frontend/packages/@depmap/wide-table/src/ReactTableV7.tsx
@@ -166,11 +166,13 @@ const ReactTableV7 = React.forwardRef(
           let newSelections: Set<any>;
           if (singleSelectionMode) {
             // We should always be getting the first selected label here since we expect selectedLabels size to be 1 in single selection mode
-            if (!prevSelections.has([...selectedLabels][0])) {
-              newSelections = new Set([...selectedLabels][0]);
+            const label = [...selectedLabels][0];
+            if (!prevSelections.has(label)) {
+              newSelections = new Set([label]);
+            } else {
+              // in the case of single selection mode we should expect prevSelections size to be 1. Effectively no change to selections here
+              newSelections = new Set(prevSelections);
             }
-            // in the case of single selection mode we should expect prevSelections size to be 1. Effectively no change to selections here
-            newSelections = new Set(prevSelections);
           } else {
             newSelections = new Set(selectedLabels);
           }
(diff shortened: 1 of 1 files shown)


frontend/packages/@depmap/wide-table/src/ReactTableV7.tsx
```diff
@@ -163,18 +163,23 @@ const ReactTableV7 = React.forwardRef(
...
+          let newSelections: Set<any>;
+          if (singleSelectionMode) {
+            // We should always be getting the first selected label here since we expect selectedLabels size to be 1 in single selection mode
+            if (!prevSelections.has([...selectedLabels][0])) {
+              newSelections = new Set([...selectedLabels][0]);
+            }
+            // in the case of single selection mode we should expect prevSelections size to be 1. Effectively no change to selections here
+            newSelections = new Set(prevSelections);
+          } else {
+            newSelections = new Set(selectedLabels);
```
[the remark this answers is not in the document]
alimourey: What new behavior do you need that this refactoring is supposed to support?

frontend/packages/@depmap/wide-table/src/ReactTableV7.tsx
```diff
@@ -163,18 +163,23 @@ const ReactTableV7 = React.forwardRef(
...
-            newSelections.clear();
-            newSelections.add([...selectedLabels][0]);
+          let newSelections: Set<any>;
+          if (singleSelectionMode) {
+            // We should always be getting the first selected label here since we expect selectedLabels size to be 1 in single selection mode
+            if (!prevSelections.has([...selectedLabels][0])) {
+              newSelections = new Set([...selectedLabels][0]);
+            }
+            // in the case of single selection mode we should expect prevSelections size to be 1. Effectively no change to selections here
+            newSelections = new Set(prevSelections);
```
alimourey, line 173: Did you mean to put `newSelections = new Set(prevSelections);` in an else block? The way it is now, it looks like this line will always overwrite the value of newSelections from the previous if block

frontend/packages/@depmap/wide-table/src/ReactTableV7.tsx
```diff
@@ -163,18 +163,23 @@ const ReactTableV7 = React.forwardRef(
...
         setSelections((prevSelections) => {
-          const newSelections = new Set(prevSelections);
-          if (!newSelections.has([...selectedLabels][0])) {
-            newSelections.clear();
-            newSelections.add([...selectedLabels][0]);
+          let newSelections: Set<any>;
+          if (singleSelectionMode) {
+            // We should always be getting the first selected label here since we expect selectedLabels size to be 1 in single selection mode
+            if (!prevSelections.has([...selectedLabels][0])) {
+              newSelections = new Set([...selectedLabels][0]);
```
alimourey, line 170: I think this should be `newSelections = new Set(selectedLabels);` since selectedLabels is already an array

frontend/packages/@depmap/wide-table/src/ReactTableV7.tsx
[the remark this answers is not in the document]
alimourey: I _think_ after the changes I mentioned in my other 2 comments, this will maintain the behavior from before your refactor. But I did notice something I'm curious about. It looks like most other parts of this file that deal with newSelections set the value of new selections as `const nextSelections = new Set(prevSelections)`, and then proceed to add/delete from that set. Whereas, this refactor declares newSelections with `let` and redefines as needed.

@rcreasi I think you wrote most of this file. Is it better to add/delete from a const set? Or is this just a matter of personal preference?

```
   const handleClickCheckbox = useCallback(
      (idValue: string) => {
        setSelections((prevSelections: Set<string>) => {
          const nextSelections = new Set(prevSelections);

          if (nextSelections.has(idValue)) {
            nextSelections.delete(idValue);
          } else {
            nextSelections.add(idValue);
          }

          setTimeout(() => {
            if (idProp && onChangeSelections) {
              onChangeSelections(
                toVisibleSelections(nextSelections, rows, idProp)
              );
            }
          }, 0);

          return nextSelections;
        });
      },
      [rows, idProp, onChangeSelections]
    );

```

frontend/packages/@depmap/wide-table/src/ReactTableV7.tsx
jessica-cheng: Ah good catc! I think I meant to se this to just the first selectedLabel since I think that's what it was doing before. I'll change that!

frontend/packages/@depmap/wide-table/src/ReactTableV7.tsx
jessica-cheng: Yes!

jessica-cheng (commented):
@alimourey Thanks for catching some potential error in my last minute refactoring changes 😅 I addressed those issues now I believe

alimourey (approved):
Looks good!

rcreasi:
I don't think there should be a `selectedLabels` prop.  How do you keep track of what's selected if there's a prop that stomps out the state? Only an `initiallySelectedLabels` prop would make sense.

Or we could make it so the table doesn't keep track of the selected state itself. But then the parent component would have to handle things like select all and that's weird. I also don't really remember what "visible selections" is all about. I think that's so you don't accidentally perform actions on hidden (filtered out) rows?

jessica-cheng:
> I don't think there should be a `selectedLabels` prop. How do you keep track of what's selected if there's a prop that stomps out the state? Only an `initiallySelectedLabels` prop would make sense.
> 
> Or we could make it so the table doesn't keep track of the selected state itself. But then the parent component would have to handle things like Select All and that's weird. I also don't really remember what "visible selections" is all about. I think that's so you don't accidentally perform actions on hidden (filtered out) rows?

@rcreasi I agree that this prop is confusing and I wasn't aware of it before until I needed a way for my parent component to handle what actually gets selected in the table. Maybe we can determine who actually controls selection state later but for now I just want to make sure that this change wouldn't break other components that use this prop currently and be unblocked for Correlations Analysis

frontend/packages/@depmap/wide-table/src/ReactTableV7.tsx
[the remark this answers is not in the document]
alimourey: That's a cool feature! I looked into it more, and I think Celligners use of selectedTableLabels was technically a mistake, because it looks like it wouldn't have actually done anything. But it wasn't noticed because it didn't cause any buggy behavior.

In Celligner, the workflow would've been:
1. You click a row, "A". 
2. The table's selections state gets updated in handleClickCheckbox to be equal to "A".
3. The graphs annotations state gets updated to "A".
4. The selectedTableLabels prop for WideTable gets updated to "A"
5. The block of code you were looking at sees prevSelections is A and sets new selections to A... In other words, it never actually did anything for Celligner.

i.e. The confusing if block you were looking at would've never been hit:

          if (!newSelections.has([...selectedLabels][0])) {
            newSelections.clear();
            newSelections.add([...selectedLabels][0]);
          }

PR #275, opened 2025-05-16
feat(breadbox): Error reporting from celery worker

snwessel:
**The high-level changes here are:**
1. Celery now has its own `GCPExceptionReporter` instance and client. This means that the celery workers will be able to report exceptions directly, instead of serializing them as a "task.result" which was dropping the stacktrace before breadbox could report them to GCP. 
2. Failed celery tasks are now returned by breadbox with 200 statuses. They will have the `TaskState.FAILURE` though and will have messages that can be displayed to users. 

**A couple of implementation notes:**
1. I did not update anything related to the portal-backend celery worker. It seemed like too much effort to be worthwhile when we'll soon be removing it anyway.
1. Reports from within celery will not have information about the request. I think this is fine - the stack trace seems much more useful for debugging. 

**Motivation:** 
* Errors from celery previously appeared in the GCP logs like this:
    image

* Each celery log will now have a full stack trace like:
```
Traceback (most recent call last):
  File "/Users/swessel/Library/Caches/pypoetry/virtualenvs/breadbox-tVRrHl2Y-py3.9/lib/python3.9/site-packages/celery/app/trace.py", line 453, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/Users/swessel/Library/Caches/pypoetry/virtualenvs/breadbox-tVRrHl2Y-py3.9/lib/python3.9/site-packages/celery/app/trace.py", line 736, in __protected_call__
    return self.run(*args, **kwargs)
  File "/Users/swessel/Desktop/Repos/depmap-portal/breadbox/breadbox/compute/analysis_tasks.py", line 300, in run_custom_analysis
    raise Exception("TEST")
Exception: TEST
```


breadbox/breadbox/logging.py
```diff
@@ -0,0 +1,73 @@
...
+        breadbox_http_context = error_reporting.HTTPContext(
+            method=request.scope["method"],
+            url=request.scope["root_path"] + request.scope["path"],
+            response_status_code=status_code,
+        )
+        return breadbox_http_context
+
+    def report(self, request: Optional[Request], status_code: Optional[int], user: Optional[str]):
+        if self.client is None:
+            print("Error reporting disabled")
```
snwessel, line 63: So far, I've just been running this in debug mode and logging the stacktrace. On monday, I'll try these changes out in IQA to make sure things are working with all of the real configurations.

snwessel:
I just tried these changes out in my test-perf environment and they're working! 💃 
image

You can see the error page here: https://console.cloud.google.com/errors/detail/CO3Yg6-Hx5jTfA;filter=%5B%5D;time=P30D;locations=global?inv=1&invt=Abx07w&project=cds-logging

breadbox/breadbox/compute/celery.py
```diff
@@ -1,13 +1,29 @@
...
+
+class LogErrorsTask(Task):
+    def on_failure(self, exc, task_id, args, kwargs, einfo):
+        """Override the default behavior on task failure to also report to our GCP error logs.""" 
+
+        # It would be nice if we could report some of the information we do have here
+        # (like task_id, args, etc.) to error_reporter. However, error_reporter is really 
+        # only set up to log context in the format of a HTTP request. 
+        exception_reporter.report(request=None, status_code=None, user=None)
+        super(LogErrorsTask, self).on_failure(exc, task_id, args, kwargs, einfo)
```
pgm, line 22: just as a fyi, the above looks like how calling a superclass's method worked in python 2. It should work but in python 3+ you can just write:
```
        super().on_failure(exc, task_id, args, kwargs, einfo)
  ```
https://docs.python.org/3.7/library/functions.html#super

File: breadbox/breadbox/startup.py [hunk 3]
```diff
@@ -96,31 +94,3 @@ async def __call__(self, scope, receive, send):
         await self.app(scope, receive, send)
 
 
-class GCPExceptionReporter:
```
pgm, line 99: Weird, we had a GCPExceptionReporter in startup.py already? I guess this wasn't being used?

pgm (approved):
LGTM

breadbox/breadbox/startup.py
snwessel: Sorry for the confusion, this definition was being used before, I just moved it to a different file now.

This `GCPExceptionReporter` class had previously been located in `startup.py`, which I tend to think of as being specific to the breadbox app - so it felt a little more natural to me to put it in with the logging utilities, now that it's being used by the background worker as well. I'm not sure it matters much to me one way or another though, so let me know if you have an opinion.

breadbox/breadbox/celery_task/utils.py
```diff
@@ -111,28 +111,23 @@ def format_task_status(task):
...
-            #     "detail": str(task.result.status_code),
-            # }
+            message = {
+                "status_code": str(task.result.status_code),
+                "detail": str(task.result.status_code),
+            }
         elif isinstance(task.result, FileValidationError):
-            raise task.result
-            # message = str(task.result)
+            message = str(task.result)
```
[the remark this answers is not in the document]
snwessel: Good point, that change might impact both the breadbox loader and the Data Manager UI. I'll give them both a try with these changes 👍

jessica-cheng (approved):
LGTM!

breadbox/breadbox/compute/celery.py
```diff
@@ -1,13 +1,29 @@
...
 
+from breadbox.logging import GCPExceptionReporter
 
-log = getLogger(__name__)
 
 rhost = os.getenv("REDIS_HOST", "localhost")
+breadbox_env = os.getenv("BREADBOX_ENV", "dev")
+
+log = getLogger(__name__)
+exception_reporter = GCPExceptionReporter(breadbox_env=breadbox_env)
```
jessica-cheng, line 12: Could we name this maybe `{breadbox_env} celery` so it's easier to distinguish which errors are coming from the celery workers vs the breadbox service itself?

breadbox/breadbox/celery_task/utils.py
```diff
@@ -111,28 +111,23 @@ def format_task_status(task):
...
             task._state = TaskState.FAILURE.name
             if isinstance(task._result, UserError):
-                message = str(task.result.detail)
+                message = str(task._result.detail)
         if isinstance(task.result, UserError):
             # this is a specific, expected error that we check for
             # return error message for the front to display
             message = str(task.result.detail)
         elif isinstance(task.result, HTTPException):
-            raise task.result
```
jessica-cheng, line 120: I'm kind of confused why we were raising `task.result` before 🤔 . I wonder if we had raised the HTTPException` itself we could have gotten around this issue of needing a separate error reporter for celery though I suppose it's helpful to distinguish which errors are coming from celery vs breadbox itself

breadbox/breadbox/celery_task/utils.py
```diff
@@ -111,28 +111,23 @@ def format_task_status(task):
             # sometimes we get an error while the celery state is still progress
             task._state = TaskState.FAILURE.name
             if isinstance(task._result, UserError):
-                message = str(task.result.detail)
+                message = str(task._result.detail)
```
jessica-cheng, line 114: @snwessel do you know what the difference is between this block of code vs the later check in line 115?

snwessel:
Looks like the Dataset Manager UI and the breadbox loader do still work well with these task failures 👍 

However, while I was testing that, I realized that even user errors were [being reported to GCP](https://console.cloud.google.com/errors/detail/CNOg2PzEzIuezgE;locations=global;time=P30D?project=cds-logging&utm_source=error-reporting-notification&utm_medium=email&utm_content=new-error&inv=1&invt=Abx1tg) when they originate from Celery (we usually filter out all 400 errors). I just updated that to follow the same rules as our other error reporters - of only reporting unexpected errors.

breadbox/breadbox/compute/celery.py
snwessel: Sure 👍 I'd been kind of on the fence about that. I'll update it to say celery in the service name

PR #278, opened 2025-05-20
refactor(breadbox): Python dependencies breadbox

naquib314:
This PR makes breadbox compatible with:
- python 3.13
- pandas 2.0
- SQLAlchemy 2.0


breadbox/breadbox/db/session.py
```diff
@@ -88,11 +90,16 @@ def SessionLocalWithUser(user: str) -> SessionWithUser:
...
+        connect_args={"check_same_thread": False},
+        future=True,
     )
 
     l = sessionmaker(
-        autocommit=False, autoflush=False, bind=engine, class_=SessionWithUser
+        autoflush=False,
+        bind=engine,
+        class_=SessionWithUser,
+        future=True,  # In SQLAlchemy 2.0, autocommit is deprecated
```
pgm, line 102: Why do we have this `future` flag? The docs say:

This package includes a relatively small number of transitional elements to allow “2.0 mode” to take place within SQLAlchemy 1.4. The primary objects provided here are [Engine](https://docs.sqlalchemy.org/en/14/core/future.html#sqlalchemy.future.Engine) and [Connection](https://docs.sqlalchemy.org/en/14/core/future.html#sqlalchemy.future.Connection), which are both subclasses of the existing [Engine](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Engine) and [Connection](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Connection) objects with essentially a smaller set of methods and the removal of “autocommit”.

Within the 1.4 series, the “2.0” style of engines and connections is enabled by passing the [create_engine.future](https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy.create_engine.params.future) flag to [create_engine()](https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy.create_engine):

That sounds to me like we shouldn't need the flag because we are using v2+, correct?

PR #279, opened 2025-05-20
Peddep landing page

jessica-cheng:
Part 1


frontend/packages/@depmap/api/src/ApiContext.tsx
```diff
@@ -77,6 +78,10 @@ export interface SharedApi {
     datasetId: string,
     datasetToUpdate: DatasetUpdateArgs
   ) => Promise<BreadboxDataset>;
+  getTabularDatasetData: (
+    datasetId: string,
+    args: TabularDatasetDataArgs
+  ) => Promise<{ [key: string]: { [key: string]: any } }>;
```
rcreasi, line 84: Looks like the same function you added on the correlations analysis branch, yeah? Like I said, I'm getting ready to replace all of this API stuff but I don't want to cause an onslaught of terrible merge conflicts. Once this gets merged, I'll rework this change into the new format. Then I'll give an overview of the changes at an upcoming TPP.

PR #280, opened 2025-05-22
New peddep landing page graph

jessica-cheng:
See graph here: https://app.asana.com/1/9513920295503/task/1210355082850369/comment/1210355082850376?focus=true


jessica-cheng (commented):
@rcreasi I haven't created a new Plotly graph from scratch in a while and this is one of the first few times we're querying directly from Breadbox. I've noticed you have been writing a lot of custom hooks lately to fetch data from Breadbox and I was wondering if you can give a review on best ways to organize such code moving forward.

frontend/packages/portal-frontend/src/peddepLandingPage/components/SubgroupsPlot.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const [bapi] = useState(() => getApi());
+
+  const [data, setData] = React.useState<SubgroupSubtypes | null>(null);
+  const [hasError, setHasError] = useState(false);
+
+  useEffect(() => {
+    (async () => {
+      try {
+        // Get depmap models data
+        const dimensionType = await bapi.getDimensionType("depmap with peddep");
```
jessica-cheng, line 21: TBD: Change to `depmap_model` after confirming the new column names

rcreasi:
> @rcreasi I haven't created a new Plotly graph from scratch in a while and this is one of the first few times we're querying directly from Breadbox. I've noticed you have been writing a lot of custom hooks lately to fetch data from Breadbox and I was wondering if you can give a review on best ways to organize such code moving forward.

What you have looks pretty good. I tend to add custom hooks when I feel like things are getting too cluttered. That's really just a matter of taste. But it also results in a fair amount of boilerplate code which isn't fun. I think we should start using [something like this instead](https://tanstack.com/query/latest/docs/framework/react/overview). I'm planning on setting that up soon. Stay tuned :tv:

rcreasi (approved):

PR #288, opened 2025-06-03
Add PediatricModelType to context manager

rcreasi:


portal-backend/depmap/data_explorer_2/datatypes.py
```diff
@@ -246,6 +246,10 @@
             "name": "PatientTumorGrade",
             "valueType": "categorical",
         },
+        "slice/depmap_model_metadata/PediatricModelType/label": {
+            "name": "PediatricModelType",
+            "valueType": "categorical",
+        },
```
rcreasi, line 252: This hardcoded list was always a bit of a hack but it now it feels extra hacky because we haven't actually added this column to all the environments yet 😱 I'm assuming that's a temporary situation so I'm not too worried about it. But I could rework this to pull the column names directly from Breadbox instead.

rcreasi:
> @rcreasi I would recommend either cherry picking or changing the base branch to `external-25q2` since that metadata column currently only exists in `external-25q2` breadbox config

I guess we could do that. But there's no way to keep track of when this change is ready for the main branch so I'm not doing that. I'll just make a little tweak.

portal-backend/depmap/data_explorer_2/datatypes.py
jessica-cheng: Ah right now I only added the column in depmap-deploy `external-25q2` branch. I figured since it was only relevant to public for now, it'll eventually get merged back into master after the public release anyways. But if we want this option to be visible in the next Skyros release I can cherry pick that config change into both `dmc-25q2` and `master`

jessica-cheng: Update: @rcreasi I've added the column to our other branches so you should be good to go now. @pgm  I'll also add this commit into `qa` so it will appear in our next Skyros release

PR #290, opened 2025-06-04
Added drc_datset_label to DoseResponseCurve

pgm:
These changes are to assist with the creation of the new dose response tab: A set of changes to allows us to find the set of DoseResponseCurves associated with a particular viability dataset.

I created drc_compound_datasets so we could match the `drc_dataset_label` with the corresponding viability dataset and also have a meaningful label to show users.

This lead me to discover that the repurposing screen was getting loaded without a `label` on the artifact, so I also added that.

Currently the DRC are all being loaded indexed by `CompoundExperiment` and not `Compound` but that is something we can tackle in the future. For the time being we can build out the new DRC page using oncref which has a 1:1 relationship between CompoundExperiment and Compound.

### DIFF
--- a/pipeline/reformat_repurposing_data.conseq
+++ b/pipeline/reformat_repurposing_data.conseq
@@ -44,6 +44,7 @@ rule reformat_curve_params_secondary_repurposing_data:
         taiga_token={"type": "config-file", "name": "taiga-token"}
     outputs: {
         "type": 'dose-response-curve-params',
+        "label": "repurposing_per_curve",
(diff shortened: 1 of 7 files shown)


pgm:
@alimourey , these changes are included in your PR https://github.com/broadinstitute/depmap-portal/pull/296 correct?

If so, I'll just close this PR because they can get merged in with your other changes.

alimourey:
> @alimourey , these changes are included in your PR #296 correct?
> 
> If so, I'll just close this PR because they can get merged in with your other changes.

Correct. I will close this.

PR #294, opened 2025-06-10
fix(breadbox): Make dimension type optional for associated datasets

jessica-cheng:
I encountered an error when trying to fetch associations for OncRef dataset. After looking into it, it seems like we are making assumptions that all associated datasets will have a feature type. This assumption breaks for the associated dataset "Subtype Matrix" since its feature type is `None`. 

@pgm I'm not too familiar with the original intention of your associations implementation and this particular dataset (It does feel a bit strange to say a compound is associated with `None`) but my following changes helped to unblock me

### DIFF
--- a/breadbox/breadbox/schemas/associations.py
+++ b/breadbox/breadbox/schemas/associations.py
@@ -1,11 +1,11 @@
 from pydantic import BaseModel
-from typing import List
+from typing import List, Optional
 
 
 class DatasetSummary(BaseModel):
     id: str
     name: str
-    dimension_type: str
+    dimension_type: Optional[str] = None
     dataset_id: str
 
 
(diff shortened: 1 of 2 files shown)


pgm (approved):
...ah! I had said LGTM ...but I misread what was going on here. I'll propose a small change.

pgm:
> @pgm I'm not too familiar with the original intention of your associations implementation and this particular dataset (It does feel a bit strange to say a compound is associated with `None`) but my following changes helped to unblock me

The association isn't between the feature type, but the measurements for the given feature. When we have a null feature type, it's intended to mean, "this matrix does not have a 'defined' feature type" or in other words, the features are unique to this dataset and have no additional data beyond the feature labels.

Does that explanation help?

PR #297, opened 2025-06-13
Get rid of little minor tick numbers on old Dose Curves page

alimourey:
Specified dtick = 1 so that ticks are set every 10^(n"dtick) where n is the tick number.
https://plot.ly/javascript/reference/#layout-xaxis-dtick 

Example of what we don't want (the little 2 and 5 numbers):
Screenshot 2025-06-13 at 4 15 03 PM


naquib314:
LGTM

PR #315, opened 2025-07-01
feat(breadbox): Added checks for ensuring that group entries always match full email or email domain.

pgm:
In `group_entry` the `exact_match == False` allows us to provide a domain name so that all users with that domain email address have access to the group. However, the check is looser than I'm comfortable with and in its previous form just checks the email address suffix.

These changes require that the suffix that's checked starts with a "@" so that it's always checking the _entire_ domain part of the email address.

### DIFF
--- a/breadbox/breadbox/api/groups.py
+++ b/breadbox/breadbox/api/groups.py
@@ -156,6 +156,12 @@ def add_group_entry(
                 raise HTTPException(404)
             raise HTTPException(403)
 
+        if not group_entry.exact_match and not group_entry.email.startswith("@"):
+            raise HTTPException(
+                422,
+                "If exact_match is False, which indicates the entry applies to all email addresses within a domain, the email field must start with a '@'",
+            )
+
         group_entry_db = group_crud.add_group_entry(db, user, group, group_entry)
     return group_entry_db
 
--- a/breadbox/breadbox/crud/access_control.py
+++ b/breadbox/breadbox/crud/access_control.py
@@ -23,6 +23,14 @@ def user_has_access_to_group(
     if user in settings.admin_users:
         return True
 
+    # adding a special case for special groups, to avoid running into the assertion below
+    # that all email entries must start with a "@" if group_entry.exact_match == False
+    # We should clean up that entry because I think it'd be better if it doesn't exist -- but for now, let's avoid the data migration.
+    # This means that only the public group which is possible to be viewable by _all_ users. Similarly all users need to be
+    # able to access the transient group.
+    if group.id in [PUBLIC_GROUP_ID, TRANSIENT_GROUP_ID] and write_access is False:
+        return True
+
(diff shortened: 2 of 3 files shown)


snwessel:
Fyi, I also just double checked that the script I used to migrate groups from our old private dataset system did append the @ to the beginning of the email domains when it added them to breadbox ([here](https://github.com/broadinstitute/depmap-deploy/blob/661ee96863fc6274e6c1facfb9e53e3c68cbf422/breadbox_migration_scripts/copy_private_groups_and_datasets.py#L51)), so our existing private group definitions should be compliant with this new restriction.

breadbox/breadbox/api/groups.py
```diff
@@ -156,6 +156,12 @@ def add_group_entry(
                 raise HTTPException(404)
             raise HTTPException(403)
 
+        if not group_entry.exact_match and not group_entry.email.startswith("@"):
```
jessica-cheng, line 159: This is just a personal preference but for validations related to the request body, I tend to put them in the request body's Pydantic model as a validator. This has the added benefit that it will return an error earlier if the request body is malformed

jessica-cheng: NOTE: Pydantic has a special type `EmailStr` which I thought we could use for the `GroupEntryIn` model but it requires that the email be formatted as `[name]@[domain]`. However, this makes me realize that perhaps there should be more specific rules to think about such as proper formatting of the domain to include only one `@` and `.` symbol. I don't think we need a super stringent check for now since we currently control groups but something to think about in the future

breadbox/breadbox/crud/access_control.py
```diff
@@ -23,6 +23,14 @@ def user_has_access_to_group(
     if user in settings.admin_users:
         return True
 
+    # adding a special case for special groups, to avoid running into the assertion below
+    # that all email entries must start with a "@" if group_entry.exact_match == False
+    # We should clean up that entry because I think it'd be better if it doesn't exist -- but for now, let's avoid the data migration.
+    # This means that only the public group which is possible to be viewable by _all_ users. Similarly all users need to be
+    # able to access the transient group.
+    if group.id in [PUBLIC_GROUP_ID, TRANSIENT_GROUP_ID] and write_access is False:
+        return True
```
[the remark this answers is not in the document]
jessica-cheng: I don't think that the current updates made in the api layer for `add_group_entry` endpoint will break our implementation of `_populate_minimal_data` since that seems to call the crud's `add_group_entry` function. However, this makes me realize that my earlier suggestion of adding the extra validation to `GroupEntryIn` would break `_populate_minimal_data`.

I agree that we should remove those special entries for transient and public groups since they don't seem necessary and we'd no longer need to consider a case where the email could be `""` (with added length constraint on the db side for extra precaution if we desire)

PR #317, opened 2025-07-03
Heatmap tab

alimourey:
These changes are support adding a new Heatmap Tab to the Compound page.

Some of the "Files changed" include changes to the dose curves tab, because the Heatmap and Dose Curves share a "DoseViabilityTable", "CompoundPlotSelections", "compoundPagePromptForSelectionFromContext", as well as some types and utils. Heatmap and Dose curves also share the same hook for fetching the dose table data.

It would be most helpful to get more sets of eyes on everything in this folder: frontend/packages/portal-frontend/src/compound/heatmapTab

As well as everything in this folder: frontend/packages/portal-frontend/src/compound/hooks.

I also made significant changes to the Jinja template located here: portal-backend/depmap/templates/compounds/index.html --> This controls the rendering of the Compound tabs. I made some changes so that the individual tabs will not load until the tab has been clicked on. This reduces the number of simultaneous requests bombarding Breadbox, thereby significantly increasing the performance of both the Dose Curves and Heatmap tabs (think 2-4 seconds to load+render vs 15-25 seconds before).

**Important Features for the Heatmap:**
- The Heatmap displays sorted compound dose viability data.
- The data starts zoomed out, but there is a draggable resizeSlider below the plot that lets the user explore the data and zoom in/out
- There is also a "Zoom to Selection" button. On click, this button will zoom to whatever resolution is the most zoomed in that will allow for all selected columns to be visible. I originally made this automatic, but it got kind of messy to have the window jumping around when the user might not want it to.

**NOTE**: I am still refining the tick mark logic that displays either a cell line name or "(number of lines selected)".

See this Asana comment for a recording of the Heatmap in action: https://app.asana.com/1/9513920295503/task/1210253218582524/comment/1210710647681141?focus=true
**_UPDATE - corrected video - the previous video showed incorrect plot units:_** https://app.asana.com/1/9513920295503/task/1210253218582524/comment/1210710647681145?focus=true


frontend/packages/portal-frontend/src/compound/heatmapTab/doseViabilityHeatmap/components/PrototypeBrushableHeatmap/customizeDragLayer.ts
```diff
@@ -0,0 +1,179 @@
...
+  onMouseOut,
+  onSelectColumnRange,
+  onChangeInProgressSelection,
+  onClearSelection,
+}: Props) {
+  if ((plot as any).alreadyConfigured) {
+    return;
+  }
+
+  hideStandardPlotlyZoomHandles(plot);
```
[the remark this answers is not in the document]
alimourey: I thought the customizeDragLayer was important for setting the column range on click and drag? 

The click and drag zoom works well without this, but there didn't seem to be a way to click and drag to select multiple columns without using your implementation. I am still experimenting so it's possible I'm missing something...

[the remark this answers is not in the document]
alimourey: Oh okay, I will test it out without `hideStandardPlotlyZoomHandles()`

frontend/packages/portal-frontend/src/compound/heatmapTab/HeatmapTabMainContent.tsx
```diff
@@ -0,0 +1,223 @@
...
+                  rel="noreferrer"
+                  style={{ textDecoration: "underline" }}
+                >
+                  {displayName}
+                  <span
+                    className="glyphicon glyphicon-new-window"
+                    style={{
+                      paddingLeft: "3px",
+                      fontWeight: 300,
+                      fontSize: "10px",
```
snwessel, line 103: I personally prefer to have all the CSS defined in classes an SCSS file as opposed to having them defined inline in the react here. _But_, our codebase isn't super consistent about that, so I'd be curious what other people think. I have worked with people in the past who feel very strongly about avoiding inline styles. Personally, I don't feel that strongly about it, but I think it can be helpful to move all of the CSS into classes in a separate file for a couple of reasons:

1. It's nice to have _all_ of the custom CSS for a component in one place. For example, if I'm trying to debug some styling issues for a tooltip or something, I can see all of the custom CSS in one place (and not need to read through a multiple large `.tsx` files). This also makes it easier to follow the inheritance of styles between elements. 
2. It makes it a bit easier to debug from the browser because when you inspect an element, you can see exactly what file + class the styles are originating from. 

For example, the DataExplorer2 CSS is set up the way I'm imagining would be ideal: https://github.com/broadinstitute/depmap-portal/blob/master/frontend/packages/%40depmap/data-explorer-2/src/components/DataExplorerPage/styles/DataExplorer2.scss

This does really come down to personal preference though, so feel free to weigh in if you have an opinion (and I can provide more background on CSS classes if that would be helpful).

frontend/packages/portal-frontend/src/compound/CompoundDoseViability.scss
```diff
@@ -0,0 +1,214 @@
...
+    margin-top: 20px;
+    margin-bottom: 20px;
+
+    p {
+      max-width: 780px;
+    }
+  }
+}
+
+.PlotSection {
```
snwessel, line 47: It looks like this is very similar to the existing `CompoundDoseCurves.scss` file that's used by the existing dose curves tab. Ideally it would be nice to avoid the duplication though. Are you thinking that the older `CompoundDoseCurves.scss` file would be removed in the near future if the older dose curves tab is removed?

If not, do you think it would be possible to remove some of the duplication by moving this file somewhere that it can be used by both components?

snwessel (commented):
Looks good! I really like the functionality on this new tab. I think the user interaction seems really intuitive, especially given how big and complicated the heatmap is. 

I had a few questions and suggestions about the organization of changes, but overall this looks good!

frontend/packages/portal-frontend/src/compound/heatmapTab/HeatmapTabMainContent.tsx
rcreasi: I also have a slight preference for keeping it in a `.scss` file but I don't feel too strongly about it. Inline styles are supposedly less performant (though I've never seen it actually make a difference) and can't be optimized by the bundler (Webpack will take two identical CSS classes and hash them to a single definition). The other strike against inline styles is that you can't override them without resorting to `!important` 🤮 . I often throw some inline styles on a component out of laziness then I go back later and do a pass where I convert them all.

rcreasi: One more point is that when I see `styles={}` on an element, it can be a hint that there's something dynamic happening that couldn't have been done with some static CSS. I'll scan to see what variables are being used and then realize there aren't any 🤷

portal-backend/depmap/settings/settings.py
```diff
@@ -100,7 +100,7 @@ def flagship_projects(self):
         return self.is_skyros()
 
     @property
-    def new_dose_curves_tab(self):
+    def new_compound_page_tabs(self):
```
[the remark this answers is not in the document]
alimourey: I could separate them into 2 feature flags if you think that would reduce confusion: dose_curves_tab_v2, and heatmap_tab. Or something like that...

portal-backend/depmap/templates/compounds/index.html
```diff
@@ -192,42 +179,79 @@ <h1 class="compound_name">{{ name }}</h1>
 <p style="padding: 15px 0">No data was found on this compound.</p>
 {% endif %} {% endblock %} {% block js %}
 <script>
-  // This is the only place we still use jquery-stickytabs. It should be
-  // removed after this page is converted to React because it doesn't
-  // play nice with our React Tabs component (it wants to stomp out the
-  // ?tab= query parameter).
```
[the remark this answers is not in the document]
alimourey: Good catch. I did not mean to delete this. Oops.

frontend/packages/portal-frontend/src/compound/heatmapTab/HeatmapTabMainContent.tsx
snwessel: Oh I see now, yes! Thank you for pointing that out (and sorry for the confusion)

I had been looking for a `.scss ` file corresponding to this heatmap component, and I missed your import of:
```
import styles from "../CompoundDoseViability.scss";
```

It looks like there are a couple of other small places in this file specifically where you had `style=...` , but if you have a specific reason to keep them inline, I think it makes sense to keep them there. Otherwise, it might be worth moving just those couple of things over to classes as well (just to have everything in one place)

alimourey: Yep, I saw those and moved them into the classes. Thanks!

PR #333, opened 2025-07-16
Heatmap tile

alimourey:
These changes support the new Compound Heatmap Tile. I separated the existing PrototypeBrushableHeatmap into 2 flavors: (1) InteractiveBrushableHeatmap, and (2) StaticBrushableHeatmap.

The interactive heatmap is what i shown on the Heatmap tab.

The static heatmap is what is shown on the Heatmap tile.

The heatmap tile reuses the Heatmap Tab's method of fetching data: useDoseTableDataContext.

Right now, OncRef is the only dataset that is supported.

**Another notable change:** The compound tiles have long been colored various shades of green. Vonne requested a switch to red to avoid having the heatmap on the tile clash with the rest of the page. Previously, we used a slightly different shade of color for each compound dataset. For now, I've switched everything over to the same shade of red regardless of dataset. To see screenshots of the new tile and red color scheme, see this Asana task: https://app.asana.com/1/9513920295503/project/1165651979405609/task/1210688748327955


depmap-shared/color_palette.json
```diff
@@ -4,13 +4,13 @@
...
-  "gdsc_color": "#006600",
-  "ctrp_color": "#99cc66",
-  "repurp_color": "#669933",
-  "rep1m_color": "#547335",
-  "rep_all_single_pt_color": "#55aa24",
+  "gdsc_color": "#c55252",
+  "ctrp_color": "#c55252",
+  "repurp_color": "#c55252",
+  "rep1m_color": "#c55252",
+  "rep_all_single_pt_color": "#c55252",
```
alimourey, line 11: A question for reviewers: Previously, we used a slightly different shade of green for each dataset. Do we still care about using different shades of the compound color per dataset, or is it fine to leave these all as the same red color?

portal-backend/depmap/compound/api.py
```diff
@@ -20,3 +21,28 @@ def get(self):
...
+
+
+@namespace.route("/prioritized_dataset")
+class PrioritizedDataset(Resource):
+    def get(self):
+        # TODO: Probably need to use this eventually to make sure the dataset has the compound
+        # compound_id = request.args.get("compound_id")
+
+        # TODO: How to figure out which dataset to prioritize? Right now only OncRef is supported,
+        # so don't worry about this yet.
```
snwessel, line 33: This is a good question, and it's one we're still trying to iron out to some extent (more discussion [here](https://app.asana.com/1/9513920295503/task/1210800769249818/comment/1210826445706204?focus=true) if you're interested). 

Even though it's not important for this PR, I'll just mention that since I believe you're only using breadbox datasets, I think it would be fine to order them by their "priority" field (which would essentially make Oncref the default).

frontend/packages/portal-frontend/src/compound/tiles/HeatmapTile/HeatmapTileContainer.tsx
```diff
@@ -0,0 +1,35 @@
...
+      );
+      setDataset(prioritizedDataset);
+      setIsLoading(false);
+    })();
+  }, [compoundId]);
+
+  return (
+    <DoseTableDataProvider dataset={dataset} compoundId={compoundId}>
+      <HeatmapTile compoundName={compoundName} isLoadingDataset={isLoading} />
+    </DoseTableDataProvider>
```
snwessel, line 33: I'm glad you've been able to keep reusing this `DoseTableDataProvider` you'd set up! It seems like a helpful pattern and really works well in this case!

frontend/packages/portal-frontend/src/compound/tiles/HeatmapTile/HeatmapTile.tsx
```diff
@@ -0,0 +1,119 @@
...
+                    ...sortedHeatmapFormattedData,
+                    x: sortedHeatmapFormattedData.x,
+                    y: sortedHeatmapFormattedData.y,
+                    z: sortedHeatmapFormattedData.z,
+                  }}
+                  onLoad={() => {}}
+                  xAxisTitle={"Cell Lines"}
+                  yAxisTitle={`Dose`}
+                  legendTitle={"Viability"}
+                  interactiveVersion={false}
```
snwessel, line 98: I'm really glad you were able to reuse this component without too much difficulty. And I like that it seems like it'll be possible to reuse this in the future as well if we incorporate heatmaps into other parts of the portal 😁

snwessel (approved):
This looks great to me!

PR #335, opened 2025-07-18
Heatmap tab updates

alimourey:
These changes address 3/4 of this Asana task: https://app.asana.com/1/9513920295503/project/1165651979405609/task/1210813380501926

Based on feedback from Phil and Mustafa's review of the new Heatmap Tab in iqa, I have made the following changes:

1. There was a request to include the dose curve params (hidden by default) in the dose viability table (the table appears on both the Heatmap Tab and the Dose Curves Tab). To make this change I moved useDoseCurveData out of its own custom hook and into what was previously named useDoseTableData. I renamed useDoseTableData to useDoseViabilityData to reflect that ALL of the dose viability data (table+heatmap+dose curves) comes from this hook.
2. I moved the definitions of the static dose viability columns into frontend/packages/portal-frontend/src/compound/utils.ts, with a separately defined hiddenDoseViabilityCols.
3. I changed the Plot Selections text from "Select points to populate list" to "Select cell lines to populate list".


frontend/packages/portal-frontend/src/compound/doseCurvesTab/DoseCurvesMainContent.tsx
```diff
@@ -109,24 +96,16 @@ function DoseCurvesMainContent({
...
-        Header: "Model ID",
-        maxWidth: 120,
-        minWidth: 80,
-      },
-      {
-        accessor: "auc",
-        Header: "AUC",
-        maxWidth: 120,
-        minWidth: 80,
-      },
```
snwessel, line 123: It's nice that you were able to move some of these configs to the `utils.ts` file, especially now that there are more columns 👍

snwessel (approved):
This looks good to me!

PR #340, opened 2025-07-24
Dose Curves and Heatmap Bug - Pages try to load for compounds without applicable data

alimourey:
Bug report: https://app.asana.com/1/9513920295503/project/1165651979405609/task/1210872955271938

The google logs picked up an issue with the new dose curves page. There was an error fetching replicates. This error was caused by a failure to hide the dose curves tab for compounds that do not have any applicable data.

The changes in this PR hide the dose curves tab, heatmap tab, and heatmap tile for compounds that don't have applicable data using data_access.is_valid_row. Tests were added to confirm the changes fix the bug.

While investigating this bug, I also noticed that the Heatmap plot displays a plot spinner forever if fetching data fails. I will make a follow up PR to fix this so that if fetching data fails, the Heatmap plot will display an error message instead.


snwessel (approved):
This looks good to me!

PR #343, opened 2025-07-25
Remove IC50 references from portal-backend

snwessel:
[Asana](https://app.asana.com/1/9513920295503/project/1165651979405609/task/1210880278875823?focus=true)

These changes remove all references to "IC50" from both the `portal-backend` and the `frontend` code. It seems like the AUC drug screen datasets were already being used as the default in all of the cases I've seen - and will continue to be used as the default data for those features. 

These IC50 drug screen datasets have already been removed from breadbox, and the next step will be to remove them from the pipeline as well.


alimourey (approved):
Looks good to me!

PR #344, opened 2025-07-25
Remove repurposing secondary dose

snwessel:
[Asana](https://app.asana.com/1/9513920295503/project/1202081696649732/task/1210885335707997?focus=true)

Remove portal-backend references to the `Repurposing_secondary_dose` dataset. After this change is merged, Phil is going to remove it from the pipeline as well. 

This dataset isn't meaningfully being used in the portal anymore, and is now largely redundant with the other "Repurposing secondary" datasets.


portal-backend/loader/dataset_loader/single_input_file_dependency_loader.py
```diff
@@ -50,9 +50,6 @@ def compound_dose_lookup(x):
     if DependencyEnum.is_compound_experiment_enum(dataset_enum):
         entity_lookup = compound_experiment_lookup
         entity_type = "compound_experiment"
-    elif DependencyDataset.is_dose_enum(dataset_enum):
-        entity_lookup = compound_dose_lookup
```
alimourey, line 54: I think this is the only place compound_dose_lookup is used, so you could probably delete that, too!

portal-backend/loader/dataset_loader/single_input_file_dependency_loader.py
```diff
@@ -50,9 +50,6 @@ def compound_dose_lookup(x):
...
-        entity_type = "compound_dose"
```
alimourey, line 55: This makes me wonder, after deleting the repurposing secondary dose dataset, is "compound_dose" still a used/valid entity_type? And if it's not a valid entity_type, can the CompoundDose sqlite table be deleted, too?

snwessel: Oh good point, thank you! Yeah it doesn't look like it's being used at all anymore.

PR #345, opened 2025-07-25
Fix error caused by heatmap/dose curves if dataset does not exist in breadbox

alimourey:
Jessica brought to my attention that data_access.valid_row will throw an error if the dataset is not a Breadbox dataset:
`ValueError: Invalid dataset GDSC2_AUC_collapsed; not in the config`.

I realized the drc_compound_datasets are only ever Breadbox datasets, so I solved this problem by adding a Breadbox specific data_access.valid_bb_row. The goal was to return False if (1) the dataset does not exist at all in Breadbox, or (2) the compound does not exist in that particular dataset.

I'm open to suggestions if people would prefer we do this differently.


snwessel (approved):
Looks good to me! Thank you for adding that!

PR #346, opened 2025-07-28
Heatmap/Dose Curves selection behavior fine tuning

alimourey:
These changes support:
1. On change of a dataset in the left Filters panel, clear selected model ids and table rows. This was necessary because not all datasets have the same model ids, so if selection was maintained, there would be confusing inconsistencies as the user switched between datasets.
- To make this change, I had to raise the state of selectedTableRows and selectedModelIds into the parent component (DoseCurvesTab/HeatmapTab). This required pulling the state out of useDoseCurvesSelectionHandlers and passing it in as props instead.

2. Heatmap only - click or click and drag deselects if the point is already selected. This Asana task includes a recording of what deselection via the Heatmap plot looks like: https://app.asana.com/1/9513920295503/project/1165651979405609/task/1210832997294438


snwessel (approved):
This looks good to me!

alimourey:
Switching this to a draft for now. People like selection stickiness so we don't actually want to keep the changes I made to support clearing selections on dataset switching (item 1 in the PR description). I might end up closing this PR and adding the changes for item 2 to a different branch.

PR #348, opened 2025-07-29
feat(breadbox): Add upload progress updates

pgm:
We poll for task status after we've uploaded a file. However, our file uploads are now taking > 1 minute for large files. In the event that things go wrong, it's hard to tell whether it's stuck or just slow, so I'm adding progress updates like we have for other "long" tasks that run inside of celery.


breadbox/breadbox/io/hdf5_utils.py
```diff
@@ -102,8 +108,11 @@ def write_hdf5_file(
                         f"Failed to update {i}:{end_col} of hdf5 file {path} with {values}"
                     ) from e
 
+        progress.update_process_max_value(None)
+        progress.update_message("Creating indexes in database")
         create_index_dataset(f, "features", pd.Index(df_wrapper.get_column_names()))
         create_index_dataset(f, "samples", pd.Index(df_wrapper.get_index_names()))
+        progress.update_message("Complete")
```
jessica-cheng, line 115: Maybe we should put this in the finally statement?

jessica-cheng (approved):
This looks good to me though I notice that ProgressTracker is only updating messages in the code block for when our df_wrapper is`ParquetDataFrameWrapper`. I know this is mostly what we're testing but it would be nice for completeness that it updates messages in the other code path for when our df_wrapper is DataFrame

breadbox/breadbox/io/hdf5_utils.py
jessica-cheng: @rcreasi That's true it doesn't make sense for progress to be considered complete for failures. I realize we haven't been catching for failures in this outer try statement but we have in our nested try except statement which assumes the failure is only expected within there so I've added one recently.

PR #349, opened 2025-07-29
Hide enriched lineages tile error

alimourey:
These changes fix the error that resulted from the Enriched Lineages tile trying to load for compounds without any enriched lineages data. There are only enriched lineage results for Prism_oncology_AUC and Rep_all_single_pt at this time.
Asana Task: https://app.asana.com/1/9513920295503/project/1165651979405609/task/1210874172963393


portal-backend/tests/depmap/compound/test_legacy_utils.py
```diff
@@ -73,6 +74,81 @@ def test_get_compound_labels_by_experiment_labels(empty_db_mock_downloads):
...
+    # Compound 2 only has one experiment (which is in our dataset)
+    compound2 = CompoundFactory()
+    compound_experiment_2 = CompoundExperimentFactory(compound=compound2)
+    # Compound 3 has one experiment (which is not in our dataset)
+    compound3 = CompoundFactory()
+    compound_experiment_3 = CompoundExperimentFactory(compound=compound3)
+
+    matrix = MatrixFactory(
+        [compound_experiment_1A, compound_experiment_1B, compound_experiment_2,],
+        [
```
alimourey, line 102: This test is the only code I actually added/changed in this file. The diff is a bit wonky because saving this file auto formatted some of the other code I hadn't touched.

portal-backend/depmap/compound/legacy_utils.py
```diff
@@ -19,7 +20,36 @@
 
 log = logging.getLogger(__name__)
 
-def _get_deduplicated_experiment_compound_mapping(dataset_id: str) -> list[tuple[CompoundExperiment, Compound]]:
+
+# compound_exp_and_dataset is a weird pre-existing hack for finding the compound experiment. Pass it in here as a param to avoid adding too many
+# independent hacks.
+def does_legacy_dataset_exist_with_compound_experiment(
+    dataset_name: str,
+    compound_exp_and_dataset: List[Tuple[CompoundExperiment, DependencyDataset]],
```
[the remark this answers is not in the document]
snwessel: Also, I'm wondering if you could add a comment to this function explaining a bit more about what it does and why it was created? As I read through this, it wasn't immediately obvious to me what it's purpose was (until I re-read your PR description and tests). In particular, it would be helpful to call out that we sometimes have `compound_exp_and_dataset` objects which aren't using the datasets that are relevant to a particular tile - assuming I'm correctly understanding that. 

The reason I ask is that I find myself doing a lot of anthropology in this part of the codebase - wondering why something was created and whether it's now (several years in the future) safe to delete.

portal-backend/depmap/compound/legacy_utils.py
```diff
@@ -19,7 +20,36 @@
...
+):
+    dataset = DependencyDataset.get_dataset_by_name(dataset_name)
+
+    # Find the dataset we are looking for in the list of comound experiment and dataset tuples
+    compound_exp_dataset = next(
+        filter(lambda item: item[1] == dataset, compound_exp_and_dataset), None,
```
snwessel, line 34: I have a bit of trouble following this line of code because
1. `compound_exp_and_dataset`is a confusing data type (which has nothing to do with your changes)
2. and there are multiple things happening in this line of code with python functions (`next` and `filter`) that I don't see used super often. 

I'm wondering if it would be possible to just break this out into multiple lines and add some more descriptive variable names? For example, I'm a little bit confused about how the variable `compound_exp_dataset` is different from `compound_exp_and_dataset`

portal-backend/tests/depmap/compound/test_legacy_utils.py
snwessel: Thank you! These tests were really helpful to me in understanding the context here

snwessel (approved):
Looks good to me! I just had a couple of suggestions about typing and readability

portal-backend/depmap/compound/legacy_utils.py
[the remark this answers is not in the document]
alimourey: Good idea! I will explain it better in the comment. And yes you are correct. Context Explorer's preprocessing scripts compute the data required for this tile. Context Explorer only produces data for OncRef and Repurposing, but some compounds only have data for GDSC2 for example. Hence this check:
```
    show_enriched_lineages = legacy_utils.does_legacy_dataset_exist_with_compound_experiment(
        DependencyEnum.Prism_oncology_AUC.value, compound_experiment_and_datasets
    ) or legacy_utils.does_legacy_dataset_exist_with_compound_experiment(
        DependencyEnum.Rep_all_single_pt.value, compound_experiment_and_datasets
    )
```

portal-backend/depmap/compound/legacy_utils.py
alimourey: Yes, I broke it out into a for loop. I think that will help with this confusion. I also added a lot more comments incrementally in this function to explain what it is doing.

portal-backend/depmap/compound/legacy_utils.py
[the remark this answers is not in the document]
alimourey: "As a heads up, I believe this object is also sometimes None" I thought you were right at first but it actually comes from [further up in the same function that renders the template](https://github.com/broadinstitute/depmap-portal/blob/cd5178ae7eac9d3817b79e72bd560d81cd6f9937/portal-backend/depmap/compound/views/index.py#L67).

I think the line you were looking at is something specific to Context Explorer, which does handle the possibility of None.

PR #353, opened 2025-07-30
Remove Context Explorer alpha tag and fix indexing of tabs

alimourey:
In preparation for Context Explorer going public, this PR (1) removes the ALPHA tag from the Context Explorer label in the tools menu, and (2) creates a map of tab types and indexes so that the explanatory text on the left side of the page will properly update even if OncRef is available. Previously, for the tab text to update properly, there had to always be 3 tabs in a precise order as defined by the TabTypes enum, but this no longer an accurate assumption.


snwessel (approved):
Looks good to me!

PR #356, opened 2025-08-04
perf(breadbox): Handle large reads

jessica-cheng:
Throw 500 error if estimated size of dataset read is too large
See [Asana task](https://app.asana.com/1/9513920295503/project/1188486168213297/task/1210870578620003)

### DIFF
--- a/breadbox/breadbox/io/hdf5_utils.py
+++ b/breadbox/breadbox/io/hdf5_utils.py
@@ -126,20 +126,25 @@ def read_hdf5_file(
 ):
     """Return subsetted df based on provided feature and sample indexes. If either feature or sample indexes is None then return all features or samples"""
     with h5py.File(path, mode="r") as f:
+        row_len, col_len = f["data"].shape
         if feature_indexes is not None and sample_indexes is not None:
+            _validate_read_size(len(feature_indexes), len(sample_indexes))
             # Not an optimized way of subsetting data but probably fine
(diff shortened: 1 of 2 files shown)


breadbox/breadbox/io/hdf5_utils.py
```diff
@@ -155,3 +160,12 @@ def read_hdf5_file(
...
             df = df.replace({np.nan: None})
     return df
+
+
+def _validate_read_size(features_length: int, samples_length: int):
+    """
+    Raise a 500 error if estimated size of reading columns and rows exceed 1GB indicating possible memory exhaustion. 
+    TODO: We will need to handle reading large data
+    """
+    if features_length * samples_length * 8 > 1024 * 1024 * 1024:
```
pgm, line 170: How about putting `1024 * 1024 * 1024` into a constant like "MAX_HDF5_READ_IN_BYTES"?  I like doing this for two reasons:
1. it gives the value on the right a clear name
2. 

Then you can also mock the value without mocking the validation function (which allows your function to also be tested. The function is pretty trivial, so fine either way. However, I wanted to highlight that constants can be mocked just like functions)

```
monkeypatch.setattr(
        "breadbox.io.hdf5_utils.MAX_HDF5_READ_IN_BYTES", 10
    )
```

pgm (approved):
Looks good but added a comment for your consideration.

PR #358, opened 2025-08-05
Heatmap and Dose Curves: prioritize datasets in dropdown

alimourey:
These changes support sorting the Heatmap and Dose Curves dataset options by the dataset priority field. The heatmap tile will also always now load the highest priority dataset that is available for that particular compound. Previously, the tile always loaded the first dataset available in the hard coded drc_compound_datasets list, but this didn't always match the correct highest priority dataset.

These changes also add the dataset display name to the Heatmap Tile.

### DIFF
--- a/portal-backend/depmap/compound/utils.py
+++ b/portal-backend/depmap/compound/utils.py
@@ -1,9 +1,9 @@
 from depmap import data_access
+from depmap.compound import new_dose_curves_utils
 from depmap.compound.models import (
     DRCCompoundDataset,
     DRCCompoundDatasetWithNamesAndPriority,
 )
-from depmap.compound.new_dose_curves_utils import get_compound_dose_replicates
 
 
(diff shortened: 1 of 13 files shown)


portal-backend/depmap/compound/utils.py
```diff
@@ -0,0 +1,59 @@
...
+
+    return with_names_and_priority
+
+
+def dataset_exists_with_compound_in_auc_and_rep_datasets(
+    drc_dataset: DRCCompoundDataset, compound_label: str, compound_id: str
+) -> bool:
+    """ 
+    We only want to load the dose curves and heatmap if the auc and replicate datasets all exist with the compound.
+    The Heatmap relies on dose curve data to load the curve params (hidden columns by default) into the table.
```
snwessel, line 37: I'd really like to believe that we always have Viability data when we have AUC data, but that _might_ not be safe to assume so I am glad you're explicitly checking 😅

snwessel (approved):
Looks good to me! I just had one question about the sorting

PR #360, opened 2025-08-05
Fix enriched lineages plot overflows bounds on gene page

alimourey:
This change fixes an issue that causes the Enriched Lineages Tile box plots to overflow their bounds on the Gene Page.

A similar bug was already fixed for the Compound Page. The Compound Page uses stickyTabs instead of React, so the Compound Page fix did not also apply for the Gene Page. However, as part of the Compound Page fix, The Collapsible Box Plot component that is used on the Enriched Lineages Tile listens for the event `changeTab:overview`.

To fix the Gene Page bug, I dispatched an event of the same name in the TabsWithHistory component.


snwessel (approved):
Looks good to me! Thank you for adding the descriptive comment!

PR #361, opened 2025-08-05
 Change Context Explorer box plots so that y axis urls open in a new tab

alimourey:
Updated the Context Explorer box plots so that the y-axis urls open in new tabs.

For more info about the reason for this change, see this Asana task: https://app.asana.com/1/9513920295503/project/1200970789253100/task/1210811611676565


snwessel (approved):
Looks good to me too 👍

PR #363, opened 2025-08-06
Correlation analysis UI fixes 2

jessica-cheng:
- Fix selection behavior where stale fields were being accessed
- Add in zoom out when double click empty space in plot
- Modify points in plot to have lighter fill color with black border when selected similar to DE2 (See [asana comment here](https://app.asana.com/1/9513920295503/task/1210920185777330/comment/1210977889953220?focus=true))
- Filter out features that are correlated with itself causing a skewed y axis
- Round correlations to 4 decimal places in table


frontend/packages/portal-frontend/src/common/utilities/plotUtils.ts
```diff
@@ -0,0 +1,9 @@
+export const hexToRgba = (hex: string, alpha: number) => {
```
jessica-cheng, line 1: @rcreasi I've moved this function to common utilities since I feel like it could be useful for other plots but I haven't changed your code yet to use this yet

alimourey (approved):
Looks good to me!

PR #366, opened 2025-08-06
Enriched Lineage Tile and Context Explorer Box Plot performance improvements

alimourey:
These changes improve the performance of the Enriched Lineages Tile and Context Explorer Box Plots. 

Instead of filtering out insignificant models in a nested list comprehension after the query, these changes use the depmap model context association table to filter within the query and also eliminate unnecessary joins.

### DIFF
--- a/portal-backend/depmap/context/models_new.py
+++ b/portal-backend/depmap/context/models_new.py
@@ -1,5 +1,6 @@
+from dataclasses import dataclass
 from operator import and_
-from typing import Dict, List, Optional, Tuple
+from typing import Dict, List, Optional, Set, Tuple
 from depmap.database import (
     Column,
     ForeignKey,
@@ -16,6 +17,12 @@ from typing import Type
 from sqlalchemy import and_, or_
 from depmap.entity.models import Entity
 from depmap.cell_line.models_new import DepmapModel, depmap_model_context_association
+
+
+@dataclass
+class OtherModelDisplayNames:
+    heme: Dict[str, str]
+    solid: Dict[str, str]
 
 
 class TreeType(enum.Enum):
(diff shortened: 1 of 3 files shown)


portal-backend/depmap/context/models_new.py
```diff
@@ -305,10 +305,27 @@ def get_model_ids_for_other_solid_contexts(
         tree_type: str,
         all_sig_models: List[str],
     ) -> Dict[str, str]:
-        contexts = (
-            db.session.query(SubtypeContext)
-            .filter(SubtypeContext.subtype_code.notin_(subtype_codes_to_filter_out))
-            .join(SubtypeNode, SubtypeNode.subtype_code == SubtypeContext.subtype_code)
+        subtype_codes_to_filter_out_set = set(subtype_codes_to_filter_out)
```
jessica-cheng, line 308: It could be better for the parameter to take in a set from the beginning since I noticed you're calling `get_model_ids_for_other_solid_contexts` and `get_model_ids_for_other_heme_contexts` one after the other meaning you're doing this set conversion multiple times

portal-backend/depmap/context/models_new.py
```diff
@@ -319,18 +336,14 @@ def get_model_ids_for_other_solid_contexts(
...
-        if len(contexts) == 0:
+        if len(depmap_model_contexts) == 0:
             return {}
 
-        model_ids = [
-            cell_line.model_id
-            for context in contexts
-            for cell_line in context.depmap_model
-            if cell_line.model_id not in all_sig_models
-        ]
```
jessica-cheng, line 330: This seems like a nice improvement!

portal-backend/depmap/context/models_new.py
```diff
@@ -348,10 +361,27 @@ def get_model_ids_for_other_heme_contexts(
         # have a signficant leaf node. Hence, the filter: SubtypeNode.level_0.notin_(subtype_codes_to_filter_out).
         # For example, if a Myeloid subtype is selected, and Lymph is signficant, we should not
         # get any "other heme" contexts from this query.
-        contexts = (
```
jessica-cheng, line 351: This is just my opinion since it seems like we could combine this into a single function since it looks very similar to the one above and I feel like we're making the same queries twice. We could maybe do final filter clause for `level_0` afterwards. I feel like this could potentially be an additional performance improvement

jessica-cheng (approved):
I think generally looks good but I do have a couple suggestions for further performance improvements!

portal-backend/depmap/context/models_new.py
alimourey: Good idea. Thanks!

PR #369, opened 2025-08-07
Delete old compound info tile code

alimourey:
We have a new "Structure and Detail" tile that replaced the old Compound Description tile. These changes delete the old, unused code.


jessica-cheng (approved):
LGTM!

PR #370, opened 2025-08-08
docs(breadbox): Document custom analysis API

snwessel:
**Making a couple of updates, with the goal of making the Custom Analysis API easier to use:**
1. Adding documentation that will appear in the Swagger docs
2. Removing an old parameter (`query_id`) which is no longer being used in breadbox. For context, we used to use this parameter to pass in a "Node ID", which was an identifier that was used in the old `vector_catalog` endpoints. The `queryId` parameter has since been updated to take feature IDs instead of Node IDs, but it's still not being used - and I don't think we don't have any plans to use it in the future.  

After these changes, the swagger page looks like this:
image

### DIFF
--- a/breadbox/breadbox/api/compute.py
+++ b/breadbox/breadbox/api/compute.py
@@ -64,6 +64,13 @@ def compute_univariate_associations(
     user: str = Depends(get_user),
     settings: Settings = Depends(get_settings),
 ):
+    """
+    Custom analysis offers three different analysis types: "pearson", "association", and "two_class". 
+    In all 3 types, the caller must specify both: a dataset and a vector of data (which we refer to as the "query vector").
+    The query vector can be specified _either_ by providing the data (queryCellLines + queryValues) or by 
+    specifying identifiers for retrieving the data within breadbox (queryFeatureId + queryDatasetId).
+    For two class comparisons, the queryValues are the strings "in" and "out".
+    """
     utils.check_celery()
 
     resultsDirPrefix = settings.compute_results_location
--- a/breadbox/breadbox/api/compute.py
+++ b/breadbox/breadbox/api/compute.py
@@ -32,27 +32,22 @@ def _get_vector_is_dependent(
 
 def _validate_parameters(
     analysis_type: str,
-    query_node_id: Optional[str],
     query_feature_id: Optional[str],
     query_dataset_id: Optional[str],
     query_values: Optional[List[str]],
     depmap_model_ids: Optional[List[str]],
 ):
-    if query_node_id and query_feature_id and query_dataset_id:
-        raise UserError(
-            f"Query feature must be specified through either node id or feature id + dataset id, not both."
-        )
-    has_query_id = query_node_id or (query_feature_id and query_dataset_id)
+    has_query_identifiers = query_feature_id and query_dataset_id
     if (
         analysis_type == models.AnalysisType.pearson
         or analysis_type == models.AnalysisType.association
     ):
         # Values can be specified with either the query_id or query_values
-        assert (has_query_id and not query_values) or (
-            query_values and not has_query_id
+        assert (has_query_identifiers and not query_values) or (
+            query_values and not has_query_identifiers
         )
     elif analysis_type == models.AnalysisType.two_class:
-        assert not has_query_id
+        assert not has_query_identifiers
         assert depmap_model_ids
         assert query_values
     else:
@@ -85,7 +80,6 @@ def compute_univariate_associations(
     assert dataset_id is not None
     _validate_parameters(
         analysis_type=analysis_type,
-        query_node_id=computeParams.queryId,
         query_feature_id=computeParams.queryFeatureId,
         query_dataset_id=computeParams.queryDatasetId,
         query_values=query_values,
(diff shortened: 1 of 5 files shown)


breadbox/breadbox/compute/analysis_tasks.py
```diff
@@ -337,16 +336,7 @@ def run_custom_analysis(
...
-                dataset_feature_id = query_node_id
-                feature_dataset = dataset_crud.get_dataset(db, user, dataset_feature_id)
-                if feature_dataset is None:
-                    raise ResourceNotFoundError("Query dataset not found")
-                query_series = get_feature_data_slice_values(
-                    db, user, dataset_feature_id, feature_dataset, filestore_location
-                )
-            elif query_feature_id and query_dataset_id:
+            if query_feature_id and query_dataset_id:
                 # The given query Id should be the id of the feature itself
```
alimourey, line 340: I realize you didn't write this recently, but when this comment says "The given query id", do you know which id it's referring to? query_feature_id?

breadbox/breadbox/schemas/compute.py
```diff
@@ -1,20 +1,41 @@
...
-    queryCellLines: Optional[List[str]] = None
-    queryValues: Optional[List[Any]] = (
-        None  # The query may be specified with either the QueryId or QueryValues
-    )
+    analysisType: Annotated[
+        Literal["pearson", "association", "two_class"],
+        Field(description="The type of analysis to run")
+    ]
+    datasetId: Annotated[
+        str, Field(description="The given ID or UUID of the dataset to search for association")
```
alimourey, line 11: I love the addition of these annotations! I feel like this will clear up a lot of confusion. I used to get so confused about which id is the dataset being searched, and which is the data slice you're querying for...

breadbox/tests/api/test_compute.py
```diff
@@ -355,8 +353,8 @@ def test_run_custom_analysis_pearson_with_feature_ids_and_query_values(
 ):
     """
     The legacy portal makes custom analysis requests to breadbox with a slightly different format.
-    1. It uses feature ids instead of node ids (with the flag use_feature_ids=True)
-    2. It sometimes passes in pre-loaded query values instead of using the query_node_id parameter
+    Instead of using the uery_feature_id + query_dataset_id params, it sometimes passes in 
+    pre-loaded query values.
```
alimourey, line 357: I'm a little confused about what it means to pass in pre-loaded query values.

alimourey (approved):
This looks great! I think it will reduce a lot of confusion!

breadbox/breadbox/compute/analysis_tasks.py
```diff
@@ -317,7 +316,7 @@ def run_custom_analysis(
     with db_context(user) as db:
 
         # All features and feature_indices for the dataset we're searching in
-        use_feature_ids = query_node_id is None
+        use_feature_ids=True
```
jessica-cheng, line 319: I'm not too familiar with how this is implemented but what is the point of having this parameter at all if we're always setting it to true?

breadbox/breadbox/api/compute.py
```diff
@@ -69,6 +64,13 @@ def compute_univariate_associations(
     user: str = Depends(get_user),
     settings: Settings = Depends(get_settings),
 ):
+    """
+    Custom analysis offers three different analysis types: "pearson", "association", and "two_class". 
+    In all 3 types, the caller must specify both: a dataset and a vector of data (which we refer to as the "query vector").
+    The query vector can be specified _either_ by providing the data (queryCellLines + queryValues) or by
```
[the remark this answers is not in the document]
alimourey: I'm not 100% sure, but I think those parameters might be for when you want to select a subset of cell lines rather than a full slice. In the Custom Analysis UI, there's an option to either "Select all cell lines" or "Select a subset of cell lines"

breadbox/tests/api/test_compute.py
snwessel: Good point! I think what this is trying to describe is the way that the feature/"query vector" data is sometimes loaded in the `portal-backend` and _then_ passed to breadbox (instead of just loading the data from within breadbox). This is one of many things that could probably be simplified once we have the fronted talking directly to breadbox. I'll update that comment to make it more clear. 👍

snwessel: I personally find this architecture really confusing, but it's been a necessary evil during the breadbox migration. I'm hoping that we can remove a lot of this complexity soon!

breadbox/tests/api/test_compute.py
```diff
@@ -352,9 +352,14 @@ def test_run_custom_analysis_pearson_with_feature_ids_and_query_values(
...
-    Instead of using the uery_feature_id + query_dataset_id params, it sometimes passes in 
-    pre-loaded query values. 
+    Not all data exists in breadbox (yet)! And as a result, we sometimes need a subset of the custom 
+    analysis data to be loaded from the portal-backend and the other subset to be loaded in breadbox. 
+    This tests the case where the feature/"query vector" data is loaded in the portal-backend and the 
+    dataset data is loaded from breadbox. Since the "dataset data" is often large, we perform the analysis in 
+    breadbox in this case. 
+
+    Once the Custom Analysis frontend has been updated to read exclusively from breadbox, we could probably
+    drop support for this functionality.
```
alimourey, line 362: This is great! It makes things so much more clear!

breadbox/tests/api/test_compute.py
pgm: Something I think about sometimes: The **entire** portal backend worker can go away once all data is in breadbox.
No more confusing shared code between the two services. No more `./flask run_worker`

That's a lot of extra moving parts that we'll be able to get rid of as soon as we can rely on breadbox for all dat.

PR #373, opened 2025-08-13
Compound dep tile style

jessica-cheng:
Fix styling so column widths more uniform and handle long compound names such that it is truncated on display but when hovering over it, displays full name

### DIFF
--- a/frontend/packages/portal-frontend/src/compound/components/tiles/TopDatasetDependencies.tsx
+++ b/frontend/packages/portal-frontend/src/compound/components/tiles/TopDatasetDependencies.tsx
@@ -25,13 +25,15 @@ export const TopDatasetDependencies: React.FC<TopDatasetDependencyProps> = ({
   const urlPrefix = window.location.origin;
   return (
     <div>
-      <h3 style={{ fontSize: "16px" }}>{dataType}</h3>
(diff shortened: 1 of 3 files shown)


frontend/packages/portal-frontend/src/compound/components/tiles/TopDatasetDependencies.tsx
```diff
@@ -58,9 +60,11 @@ export const TopDatasetDependencies: React.FC<TopDatasetDependencyProps> = ({
                     <p style={{ paddingLeft: "12px" }} />
                   )}
                   <a
+                    className={styles.ellipsisStyle}
                     href={`${urlPrefix}/${featureType}/${datasetCor.other_dimension_label}`}
                     target="_blank"
                     rel="noopener noreferrer"
+                    title={datasetCor.other_dimension_label}
```
alimourey, line 67: I noticed recently that we have a Tooltip component that @rcreasi uses beautifully in `frontend/packages/@depmap/data-explorer-2/src/components/DataExplorerPage/components/plot/integrations/GeneTea/GeneTeaTable.tsx`

I wonder if this would be a good place to use the tooltip component instead of title 🤔

PR #379, opened 2025-08-20
simplify handling of legacy Breadbox errors

rcreasi:
@jessica-cheng I was going to suggest this to simplify the error checking a little bit. What do you think?


rcreasi:
Whoops, these are all now `ErrorTypeError`s and I should be checking for that. Will fix...

jessica-cheng (approved):
LGTM! It's a lot cleaner 🧹

PR #388, opened 2025-08-29
Update long reads to get new datasets

alimourey:
Corresponding depmap-deploy PR: https://github.com/broadinstitute/depmap-deploy/pull/99

### DIFF
--- a/pipeline/data_page/get_all_data_availability.conseq
+++ b/pipeline/data_page/get_all_data_availability.conseq
@@ -4,8 +4,7 @@
 # won't manifest. Then, use the type field on each artifact to map it to the name the script is expecting
 rule get_all_data_availability:
     inputs:
-        artifacts=all {"type" ~ "depmap_data_taiga_id|depmap_oncref_taiga_id|rna_merged_version_taiga_id|rnai_drive_taiga_id|repurposing_matrix_taiga_id|ctd2-drug-taiga-id|gdsc_drug_taiga_id|raw-rppa-matrix|proteomics-raw|sanger_methylation_taiga_id|biomarker-correctly-transposed|ccle_mirna_taiga_id|ataq_seq_taiga_id|olink_taiga_id|sanger-proteomics|depmap_paralogs_taiga_id|depmap_long_reads_gcloud_loc"},
-        depmap_long_reads_datasets = all {'type':'depmap_long_reads_dataset'},
+        artifacts=all {"type" ~ "depmap_data_taiga_id|depmap_oncref_taiga_id|rna_merged_version_taiga_id|rnai_drive_taiga_id|repurposing_matrix_taiga_id|ctd2-drug-taiga-id|gdsc_drug_taiga_id|raw-rppa-matrix|proteomics-raw|sanger_methylation_taiga_id|biomarker-correctly-transposed|ccle_mirna_taiga_id|ataq_seq_taiga_id|olink_taiga_id|sanger-proteomics|depmap_paralogs_taiga_id|depmap_long_reads_dataset"},
         rnai_broad_only={"type": "raw-dep-matrix", "label": "RNAi_Ach"},
         crispr_screen_sequence_map={"type": "crispr-screen-sequence-map"},
         script=fileref("get_all_data_availability.py"),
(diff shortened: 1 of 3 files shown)


pipeline/data_page/get_all_data_availability.py
```diff
@@ -608,12 +596,13 @@ def main(
     )
     assert omics_summary.index.is_unique
 
-    # Long Reads
+    #################
+    ### Long Reads ###
+    #################
     long_reads_summary = None
-    if depmap_long_reads_gcloud_loc is not None:
+    if len(depmap_long_reads_taiga_ids) > 0:
```
alimourey, line 603: @snwessel This should skip adding the data to the plot for the public environment. We should double check this once public is staged

pipeline/data_page/get_all_data_availability.conseq
```diff
@@ -5,6 +5,7 @@
 rule get_all_data_availability:
     inputs:
         artifacts=all {"type" ~ "depmap_data_taiga_id|depmap_oncref_taiga_id|rna_merged_version_taiga_id|rnai_drive_taiga_id|repurposing_matrix_taiga_id|ctd2-drug-taiga-id|gdsc_drug_taiga_id|raw-rppa-matrix|proteomics-raw|sanger_methylation_taiga_id|biomarker-correctly-transposed|ccle_mirna_taiga_id|ataq_seq_taiga_id|olink_taiga_id|sanger-proteomics|depmap_paralogs_taiga_id|depmap_long_reads_gcloud_loc"},
+        depmap_long_reads_datasets = all {'type':'depmap_long_reads_dataset'},
```
alimourey, line 8: @snwessel I'm a little unsure of whether I used "all" correctly here. There should be 4 artifacts of type "depmap_long_reads_dataset", and I want to get them all as a list to use in the get_all_data_availability.py script. If you have issues running this, this will be the first place to check.

pipeline/data_page/get_all_data_availability.py
```diff
@@ -66,29 +67,37 @@ def preprocess_omics_dataframe(df, dataset_id):
     4. Set ModelID as index
     5. Drop columns with all NaN values
     """
```
alimourey, line 69: @snwessel The Long Reads files now have this IsDefaulyEntryForModel file. This makes me wonder if the datasets need to go through this `preprocess_omics_dataframe` step. This is something you should probably double check with Phil before deploying these changes to prod.

pipeline/data_page/get_all_data_availability.py
```diff
@@ -413,39 +422,23 @@ def get_paralogs_summary(tc, depmap_paralogs_taiga_id):
...
-        blob_name = f"{prefix}/{file_name}"
-        print(f"blob_name: {blob_name}")
-        blob = bucket.blob(blob_name)
-        content = blob.download_as_string()
-        df = pd.read_csv(pd.io.common.BytesIO(content), usecols=[0])
-        # Currently, the column name is "ACHID". Could be changed in the future to "ModelID" or something else?
-        assert "ACHID" in df.columns, f"Column 'ACHID' not found in file {file_name}"
-        unique_model_ids.update(df["ACHID"].unique())
+    for taiga_id in taiga_ids:
+        df = tc.get(taiga_id)
```
alimourey, line 438: @snwessel If the datasets do need to use `preprocess_omics_dataframe`, it should be used here: `preprocess_omics_dataframe(df, taiga_id)`

pipeline/data_page/get_all_data_availability.py
snwessel: Oh right, that's a good point, thank you! In that case, it might be good to get @naquib314 's review on this too when he's back on Tuesday - he's probably the most familiar with how these new omics indices are handled here

pipeline/data_page/get_all_data_availability.py
```diff
@@ -66,29 +67,37 @@ def preprocess_omics_dataframe(df, dataset_id):
...
+        "OmicsProfiles",
+        "OmicsSomaticMutations",
+    ]:
         print(f"Warning: {dataset_id} has multiple entries per ModelID")
     else:
-        assert not filtered_df["ModelID"].duplicated().any(), f"Duplicate ModelID after filtering in {dataset_id}"
+        assert (
+            not filtered_df["ModelID"].duplicated().any()
+        ), f"Duplicate ModelID after filtering in {dataset_id}"
         print("Setting ModelID as index...")
```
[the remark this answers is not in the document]
rcreasi: I wish [hide whitespace](https://github.com/broadinstitute/depmap-portal/pull/388/files?diff=split&w=1) were on by default. It makes this kind of thing vanish from the diff (but you have to explicitly turn it on every time you view one).

naquib314:
Cool! The changes should work and we can get an earlier pipeline run. I will merge it in.

PR #392, opened 2025-09-03
make `evaluateLegacyContext` forward compatible 

rcreasi:
This is a deprecated function that will be removed soon anyway. There would be no real reason to call it while in "Breadbox mode." But in the off chance that someone tries to do that, it will now detect that and defer the API request to Breadbox.

### DIFF
--- a/frontend/packages/@depmap/data-explorer-2/src/services/deprecatedDataExplorerAPI/portalMethods.ts
+++ b/frontend/packages/@depmap/data-explorer-2/src/services/deprecatedDataExplorerAPI/portalMethods.ts
@@ -1,10 +1,13 @@
 import qs from "qs";
+import { cached, breadboxAPI } from "@depmap/api";
 import { ComputeResponseResult } from "@depmap/compute";
 import { getUrlPrefix } from "@depmap/globals";
 import {
   DataExplorerAnonymousContext,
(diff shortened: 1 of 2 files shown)


frontend/packages/@depmap/data-explorer-2/src/services/deprecatedDataExplorerAPI/portalMethods.ts
```diff
@@ -148,6 +151,21 @@ export async function fetchLegacyAssociations(
 export async function evaluateLegacyContext(
   context: DataExplorerContext | DataExplorerAnonymousContext
 ): Promise<string[]> {
+  // It's possible that we've been passed a V2 context
+  // because this hacky wrapper is not properly typed:
+  // https://github.com/broadinstitute/depmap-portal/blob/5b38776/frontend/packages/@depmap/data-explorer-2/src/components/ContextSelector/index.tsx#L6-L15
```
rcreasi, line 156: (always uses V1 props even when V2 is returned)
https://github.com/broadinstitute/depmap-portal/blob/5b38776201e06a104f139d12f4196a900e9f238b/frontend/packages/@depmap/data-explorer-2/src/components/ContextSelector/index.tsx#L6-L15

PR #404, opened 2025-09-24
Upload PRISM datasets to breadbox

snwessel:
[Asana](https://app.asana.com/1/9513920295503/project/1165651979405609/task/1211024926695522?focus=true)

The primary goal here is to update the existing `/interactive/from-csv-url `endpoint to upload datasets to breadbox instead of the legacy database. Breadbox does a better job of managing transient datasets expiry times, which is behavior we want for these PRISM csv uploads (to be included in this public release).

In making this change, I updated a function called `upload_transient_csv`, which is used by a couple of other endpoints, as well. Calls to these endpoints will likely be phased out in the Q4 release (and Randy is in the process of updating the frontend to call breadbox directly instead of using these):
* `/interactive/api/dataset/add-csv-one-row`
* `/interactive/api/dataset/add-csv`

This required a bit more refactoring than I expected because the PRISM uploads may be indexed by _either_ depmap_id or CCLE name (i.e. cell line display name).

**Testing these changes:**
* I wrote/updated unit tests for any data transformations (anything that actually changed a dataframe)
* I did some manual integration testing of this with both real public PRISM CSV URL and a test CSV file URL I'd pulled from our github repo ([here](https://github.com/broadinstitute/depmap-portal/blob/master/breadbox/tests/sample_data/expression.csv))
* I deleted some of the existing tests which had focused primarily on ensuring that the "dataset config" was defined correctly in the legacy database.


portal-backend/depmap/cli_commands/db_load_commands.py
```diff
@@ -1558,6 +1558,7 @@ def _sync_metadata_to_breadbox():
         dim_type = dim_type_by_name.get(dimension_type.type_name)
         if dim_type is not None:
             metadata_dataset_id = dim_type.metadata_dataset_id
+            assert metadata_dataset_id
```
snwessel, line 1561: Updating the version of the `breadbox-client` used by the `portal-backend` meant that some of the client's return types changed slightly - so I made a couple of changes like this to address new pyright errors.

portal-backend/depmap/data_access/breadbox_dao.py
```diff
@@ -45,7 +42,7 @@ def get_all_matrix_datasets() -> list[MatrixDataset]:
     """
     matrix_datasets = []
     for dataset in _get_breadbox_datasets_with_caching():
-        if dataset.format_ == MatrixDatasetResponseFormat.MATRIX_DATASET:
+        if dataset.format_ == "matrix_dataset":
```
snwessel, line 45: Similarly, I only changed this because the newer versions of the breadbox client return a different type for the dataset format field.

pgm (approved):
Seems like you went to heroic efforts to retain as much of the old code as possible.

LGTM

portal-backend/depmap/user_uploads/tasks.py
```diff
@@ -154,6 +155,10 @@ def validate_csv_format(csv_path: str, single_column: bool = False):
 
         df.columns = ["custom data"]
         df.index.name = None
+    elif not is_transpose:
```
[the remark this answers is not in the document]
snwessel: Yeah it's definitely confusing, and is a big source of confusion in the old data explorer 2 backend code as well - for now, there's easy way around it though

PR #593, opened 2026-02-12
Bug: Custom analysis run on uploaded dataset not working

rcreasi:
The underlying issue is that submitting a POST to /datasets/ is no longer supported by Breadbox.

This updates a couple of compontents (Custom Analysis and the "Plot from CSV" button on the Data Explorer landing page) to use the /dataset-v2/ endpoint instead.


pgm (approved):
That was quick! I may not be the best judge, but looks good to me!

PR #636, opened 2026-03-26
delete legacy Data Explorer code

rcreasi:
This removes all the components and code branches that only served to ease the transition from the legacy DE2 backend to Breadbox. This amounts to 12K fewer lines of code (about 2% of the whole codebase).


pgm (approved):
Very large set of changes -- but mostly deletes. I poked around for a bit and convinced myself:
1. Things look generally fine.
2. If there's anything that's not fine, it's very unlikely I'll be able to spot it reading through this diff.
I say, merge it in and we'll go from there.

PR #666, opened 2026-04-09
PR 3 — Null-guarded negated operators and `is_null`/`not_null`

rcreasi:
Fixes a latent bug where entities with missing data were incorrectly matched by negated filter operators, and adds explicit null-checking operators since the old implicit null-matching path is now closed.

Builds on PR 2.

## Changes

- **`depmap_compute_embed/context.py`**:
  - New `is_null` / `not_null` unary operators.
  - New `_resolve_complements` preprocessing pass, called from `__init__` after slice loading. Desugars `!in`, `!has_any`, `!=`, and the `complement` node into null-guarded negations (`and` of `not_null` guards plus `!` of the positive form). Applies per variable reference in the expression.
  - Supporting helpers: `_collect_vars`, `_validate_var_refs`, and the `_NEGATED_OPS` table.
- **`tests/depmap_compute_embed/test_context.py`**: adds 15 tests covering the null behavior of each operator, the `complement` node, `is_null`/`not_null`, and nested expressions.

## Behavior change (release notes)

Previously, filters using "is not", "is not in list", or "has none of" could include items with no data for the selected property. For example, filtering for cell lines where lineage "is not" Breast would include cell lines with no lineage information at all, even though we don't actually know their lineage.

Items with missing data are now excluded from these filters. When using these operators to define groups (e.g., in a two-class comparison), items with missing data will fall into the "Other" category rather than being assigned to either group. Two new operators, "has no value" and "has a value", are available for filters that need to explicitly find items with missing data.

## How to test

```
pytest breadbox/tests/depmap_compute_embed/test_context.py
```

### DIFF
--- a/breadbox/breadbox/depmap_compute_embed/context.py
+++ b/breadbox/breadbox/depmap_compute_embed/context.py
@@ -27,6 +27,23 @@ operations.update(
             if isinstance(a, list) and isinstance(b, list)
             else True
         ),
+        # Unary null-checking operators. These are necessary because the
+        # implicit complement desugaring (below) closes off the old accidental
+        # way of matching null values through negated operators.
+        "is_null": lambda a: a is None,
+        "not_null": lambda a: a is not None,
+        # Note: When used with {"var": ...} references, !in, !has_any, and !=
+        # are desugared by _resolve_complements into null-guarded negations
+        # during ContextEvaluator.__init__. This prevents null values from being
+        # incorrectly matched (e.g. null !in ["a", "b"] would otherwise be True).
+        #
+        # Note: "complement" is not an operator users write directly — it is
+        # synthesized by the UI when a user selects the "NOT My Context" version
+        # of a positive context for use as an outgroup in a two-class comparison.
+        # It is desugared by _resolve_complements into a null-guarded negation
+        # so that the synthesized outgroup respects the same data-completeness
+        # gates as the positive. See _resolve_complements for the full rules.
+        #
         # Note: {"context": "<name>"} references are not standard JsonLogic.
         # They are resolved during ContextEvaluator.__init__ by
         # _resolve_context_refs, which recursively evaluates the named
@@ -143,6 +160,9 @@ class ContextEvaluator:
 
         # Resolve { "context": "<name>" } references into flat ID lists
         self.expr = self._resolve_context_refs(self.expr, context.get("contexts", {}))
+
+        # Desugar negated operators and complement into null-guarded negations
+        self.expr = _resolve_complements(self.expr)
 
         # Validate that all var references in the expression are defined
         _validate_var_refs(self.expr, self.slice_data)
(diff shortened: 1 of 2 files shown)


breadbox/breadbox/depmap_compute_embed/context.py
```diff
@@ -272,3 +290,43 @@ def _validate_var_refs(expr, slice_data: dict):
...
+
+# Negated operators that should be desugared into null-guarded negations.
+# This prevents null values from being incorrectly matched by negated operators
+# (e.g. null !in ["Breast", "Lung"] would otherwise evaluate to True).
+_NEGATED_OPS = {"!in": "in", "!has_any": "has_any", "!=": "=="}
+
+
+def _resolve_complements(expr):
+    """
+    Desugars negated operators (!in, !has_any, !=) and explicit "complement"
```
pgm, line 337: The use of "desugars" threw me off at first (I was expecting it was doing some term rewriting between equivalent forms) and I had to think about what was going on in here. It's actually "fixing" the negated expressions to work properly with nulls. I think I'd describe this as something more like, "Added null-guards to all negations"

I have to admit this is very different way to solve the problem then I would have come up with. But I think this solves the problem and is less invasive then what I probably would have come up with.

pgm: Ah, I found the desugaring! 

```
if op in _NEGATED_OPS:
            positive_op = _NEGATED_OPS[op]
            ...
            negated = {"!": {positive_op: inner}}
```

Okay, I think I get it now.

breadbox/breadbox/depmap_compute_embed/context.py
```diff
@@ -272,3 +290,43 @@ def _validate_var_refs(expr, slice_data: dict):
...
+            positive_op = _NEGATED_OPS[op]
+            inner = _resolve_complements(expr[op])
+            vars_used = sorted(_collect_vars(inner) - {"given_id"})
+            null_guards = [{"not_null": [{"var": v}]} for v in vars_used]
+            negated = {"!": {positive_op: inner}}
+            if not null_guards:
+                return negated
+            return {"and": [*null_guards, negated]}
+        if op == "complement":
+            inner = _resolve_complements(expr["complement"])
```
pgm, line 403: Does this correctly handle something like `{"complement": {"null: [{"var": "x"}]}}`?

If I read this correctly, that would become
```
{"complement": {"and: [{"not_null": [{"var": "x"}]}, {null: [{"var": "x"}]}}]
```
(or something like that) Makes my head hurt trying to map it out -- but my point is, `is_null` is the only operator which doesn't return false when evaluated on null. Doesn't this need to watch out for that edge case? ie: You wouldn't want a null guard on a is_null, would you?

pgm (approved):
One edge case I flagged. I'm not entirely sure it's a probably. Probably making a small test case for that one scenario would confirm there's a bug or prove that case is properly handled.

breadbox/breadbox/depmap_compute_embed/context.py
rcreasi: Maybe it used the word "desugars" because `!in` and `!has_any` are operators I invented and are merely sugar for `{ "!": { "in": ... } }` and `{ "!": { "has_any": ... } }`, respectively. But it is certainly weird to call `!=` "sugar."

breadbox/breadbox/depmap_compute_embed/context.py
rcreasi: Right, you wouldn't want a null guard on `is_null` because the only point of that operator is to give the user an "escape hatch" where they can assert nullness explicitly.

breadbox/breadbox/depmap_compute_embed/context.py
rcreasi: And yes, Claude came up with all sorts of convoluted solutions until I told it to add a `"complement"` operator whose sole job is to compute all the null-safe negations.

breadbox/breadbox/depmap_compute_embed/context.py
rcreasi: I'll add a test to make sure that case is handled.

rcreasi: @pgm Plaudits from Claude

claude-1

claude-2

pgm: Sometimes I feel good about myself when Claude confirms my ideas. (And then I remember it has a tendency to always say the user is right. ;) )

Regardless, it does make me feel like I'm contributing something which is nice. All these years of programming aren't totally obsolete (yet).

PR #668, opened 2026-04-09
PR 1 — Add `reindex_through` support to `SliceQuery`

rcreasi:
Extends `SliceQuery` with an optional `reindex_through` field that describes a chain of FK joins needed to reach data from a different index type. The chain nests leaf-outward: the outermost query is the data column, each `reindex_through` level is one step closer to the caller's index type.

## Changes

- **`depmap_compute_embed/slice.py`**: adds `reindex_through: Optional["SliceQuery"] = None` to the dataclass.
- **`service/slice.py`**: adds `_flatten_reindex_chain` (with cycle detection keyed on `(dataset_id, identifier)` and a max depth of 10) and `_resolve_reindex_chain`. `get_slice_data` now walks the chain and composes the FK joins into a single Series indexed by the root entity type.
- **`schemas/context.py`**: adds the `SliceQueryRef` Pydantic model (used by the endpoint below). Deliberately scoped to just the class — `Context.vars` retains its current type, `contexts` field is not added.
- **`api/datasets.py`**: `get_dimension_data` accepts an optional `reindex_through` body param typed as `SliceQueryRef`. Helpers `_to_internal_slice_query` and `_get_root_query` handle conversion and label lookup. Label resolution uses the root query so chained results get labels for the root's dimension type.
- **`depmap_compute_embed/context.py`**: adds the `_dict_to_slice_query` recursive helper and wires it into `_JsonLogicVarLookup.__getitem__` so context vars with `reindex_through` are passed through to `get_slice_data`.
- Tests added in `tests/service/test_slice.py` (chain validation, cycle detection, depth limit, single/two-hop integration, missing FK values) and `tests/depmap_compute_embed/test_context.py` (three passthrough tests).

## How to test

```
pytest breadbox/tests/service/test_slice.py breadbox/tests/depmap_compute_embed/test_context.py
```

## Notes

Frontend counterparts (`AnnotationSelectV2`, `SliceTable`) live on a separate branch and will get their own PRs after this stack lands.

### DIFF
--- a/breadbox/breadbox/api/datasets.py
+++ b/breadbox/breadbox/api/datasets.py
@@ -57,6 +57,7 @@ from ..schemas.dataset import (
     DimensionDataResponse,
     SliceQueryIdentifierType,
 )
+from breadbox.schemas.context import SliceQueryRef
 from breadbox.service import dataset as dataset_service
 from breadbox.service import metadata as metadata_service
 from breadbox.service import slice as slice_service
@@ -67,6 +68,26 @@ from ..utils.caching import CachingCaller
 
 router = APIRouter(prefix="/datasets", tags=["datasets"])
 log = getLogger(__name__)
+
+
+def _to_internal_slice_query(ref: SliceQueryRef) -> SliceQuery:
+    """Convert a Pydantic SliceQueryRef to the internal SliceQuery dataclass."""
+    return SliceQuery(
+        dataset_id=ref.dataset_id,
+        identifier=ref.identifier,
+        identifier_type=ref.identifier_type,
+        reindex_through=_to_internal_slice_query(ref.reindex_through)
+        if ref.reindex_through
+        else None,
+    )
+
+
+def _get_root_query(sq: SliceQuery) -> SliceQuery:
+    """Chase reindex_through to find the root (innermost) step in the chain."""
+    current = sq
+    while current.reindex_through is not None:
+        current = current.reindex_through
+    return current
 
 
 @router.get(
@@ -411,7 +432,13 @@ def get_dimensions(
     response_model=DimensionDataResponse,
 )
 def get_dimension_data(
-    # The request body should be a SliceQuery with the following three fields:
+    # TODO: This endpoint inlines the SliceQuery fields as individual Body() params
+    # instead of referencing the SliceQuery schema directly. This is historical drift.
+    # We're keeping it this way for now to avoid breaking the auto-generated Breadbox
+    # Python client and the Breadbox Facade, which may depend on the Body_get_dimension_data
+    # schema name and structure. Once we've confirmed those consumers can handle the
+    # change, we should refactor this to accept a single SliceQuery (or SliceQueryRef)
+    # body parameter, which would also make the OpenAPI spec self-documenting.
     dataset_id: Annotated[str, Body(description="The UUID or given ID of a dataset.")],
     identifier: Annotated[
         str,
@@ -425,6 +452,12 @@ def get_dimension_data(
             description="Denotes the type of identifier being used and the axis being queried."
         ),
     ],
+    reindex_through: Annotated[
+        Optional[SliceQueryRef],
+        Body(
+            description="Optional chain of FK joins to reindex the result by a different dimension type."
+        ),
+    ] = None,
     db: SessionWithUser = Depends(get_db_with_user),
     settings: Settings = Depends(get_settings),
 ):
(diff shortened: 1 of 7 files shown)


breadbox/breadbox/service/slice.py
```diff
@@ -83,6 +89,92 @@ def resolve_slice_to_components(
...
+            series_chain.append(get_slice_data(db, filestore_location, simple_query))
+        except Exception as e:
+            message = (
+                f"Failed to resolve reindex_through chain at {role} "
+                f"(dataset_id={step.dataset_id!r}, "
+                f"identifier={step.identifier!r}, "
+                f"identifier_type={step.identifier_type!r}): {e}"
+            )
+            log.warning(message, exc_info=True)
+            raise UserError(message) from e
```
rcreasi, line 168: @snwessel This will now log an additional warning like

> WARNING:  | 2026-04-10 16:07:45 | Failed to resolve reindex_through chain at intermediate step 1 (dataset_id='modelcondition_metadata', identifier='BAD_COLUMN_NAME', identifier_type='column')

instead of just:

> WARNING:  | 2026-04-10 16:24:45 | In get_subsetted_tabular_dataset_df, missing columns: {'BAD_COLUMN_NAME'} for dataset: 3bfadb95-c225-44cf-a041-920dc4651497

which I find much more informative. I also included `exc_info=True` which gives the full stack trace.

breadbox/breadbox/api/datasets.py
```diff
@@ -411,7 +432,13 @@ def get_dimensions(
     response_model=DimensionDataResponse,
 )
 def get_dimension_data(
-    # The request body should be a SliceQuery with the following three fields:
+    # TODO: This endpoint inlines the SliceQuery fields as individual Body() params
```
pgm, line 435: If I understand the issue correctly, this feels like jumping through hoops to be backwards compatible (a good thing) but some of the hoops we're jumping through are a result of how the autogenerated client works with types (a bad thing).

I think this change is reasonable but this is one more vote towards changing our client generation strategy. (I'm increasingly convinced we should move to LLM based generation directly from openapi spec and skip the middle autogenerated client)

breadbox/breadbox/service/slice.py
```diff
@@ -83,6 +89,92 @@ def resolve_slice_to_components(
...
+            role = f"intermediate step {i}"
+
+        simple_query = SliceQuery(
+            dataset_id=step.dataset_id,
+            identifier=step.identifier,
+            identifier_type=step.identifier_type,
+        )
+        try:
+            series_chain.append(get_slice_data(db, filestore_location, simple_query))
+        except Exception as e:
```
pgm, line 160: Eh, ordinarily I'd prefer to not catch `Exception` because it can make implementation bugs get masked and present as user input errors ... but I bet get_slice_data doesn't cleanly handle all potential edge cases.

Probably a reasonable compromise here.

breadbox/breadbox/service/slice.py
```diff
@@ -83,6 +89,92 @@ def resolve_slice_to_components(
...
+            message = (
+                f"Failed to resolve reindex_through chain at {role} "
+                f"(dataset_id={step.dataset_id!r}, "
+                f"identifier={step.identifier!r}, "
+                f"identifier_type={step.identifier_type!r}): {e}"
+            )
+            log.warning(message, exc_info=True)
```
pgm, line 167: I see `exc_info` is passed True. Okay, I think this makes me okay with catching `Exception`. If there is an implementation error, we'll have the stacktrace in the log which we'll be able to use to investigate.

pgm (approved):

breadbox/breadbox/service/slice.py
rcreasi: It's not that it doesn't catch the edge cases (it does), it's that you lose the context that it happened as part of a chain of calls.

rcreasi: I added this just a moment ago after Sarah and I looked at it in paired code and tried to imagine the failure mode.

rcreasi: (see comment above ☝️ https://github.com/broadinstitute/depmap-portal/pull/668#discussion_r3066596300)

breadbox/breadbox/api/datasets.py
```diff
@@ -69,6 +70,26 @@
 log = getLogger(__name__)
 
 
+def _to_internal_slice_query(ref: SliceQueryRef) -> SliceQuery:
```
snwessel, line 73: I do think it would be good to circle back at some point to combine `SliceQueryRef` (the pydantic model) and `SliceQuery` (the dataclass) - probably by updating SliceQuery to use pydantic - but that can be done as a separate set of changes. Having a pydantic version of this class feels like a step in the right direction 👍 

I'm guessing the reason I'd originally defined `SliceQuery` as a dataclass was that it was defined in the `depmap-compute` module (which was shared between the `portal-backend` and `breadbox` and didn't have references to pydantic) - but that's not the case anymore. And it may be easier to update once we make the breadbox client changes Phil mentioned

snwessel (approved):
Looks good to me!

breadbox/breadbox/service/slice.py
snwessel: Yeah this catches errors from a set of pandas joins that I think have the potential to run into unexpected issues (at least it's been my experience that the more advanced pandas functionality has a tendency to run into weird edge cases), and it'll be really helpful to have some log of what data lead to any given error

rcreasi: Claude also thought it was worth logging a warning whenever an intermediate step produces all NAs. Those ultimately get dropped resulting in an empty response set. The problem with that is, at least in the case of screen pairs (which track only a small fraction of known models), that is a very frequent occurrence and our logs would get littered with such messages.

PR #697, opened 2026-06-09
Molecular subtype tree updates

acameron13:
Editing the `create_subtype_tree.py` script to take in a predetermined hierarchy of molecular subtypes, rather than rely on string parsing to construct the hierarchy. The predetermined hierarchy is now stored in taiga and is passed in as a conseq input.

At a high level, this involves using the same function for the oncotree and the molecular subtypes to go from a hierarchy to the table format. The previous function was called `create_oncotable`, and has been edited and renamed to be `create_table_from_hierarchy`. The old molecular subtype code has been deleted.

Testing on my local machine confirms that it produces the exact same output as the 26Q1 release.


snwessel (approved):
I'm not super familiar with this, but these changes look good to me!

PR #698, opened 2026-06-12
fixed mistake which caused 'self' points to be dropped from correlation analysis

pgm:
Specifically, when viewing correlation analysis results for CRISPR BRAF, the BRAF Hotspot Mutation is being dropped.

It should only filter out CRISPR BRAF correlated with CRISPR BRAF. (Because it will be 1 by definition and messes up the scale on the plots)

Problem was trivial mistake in logical test

### DIFF
--- a/frontend/packages/portal-frontend/src/correlationAnalysis/hooks/useCorrelationAnalysisData.ts
+++ b/frontend/packages/portal-frontend/src/correlationAnalysis/hooks/useCorrelationAnalysisData.ts
@@ -77,8 +77,10 @@ export function useGeneCorrelationData(
 
         const filtered = rawCorrelates.filter(
           (cor: SortedCorrelations) =>
-            cor.feature !== featureName &&
-            cor.featureDatasetGivenId !== selectedDataset.datasetId
+            !(
+              cor.feature === featureName &&
+              cor.featureDatasetGivenId === selectedDataset.datasetId
+            )
         );
 
         setState({


frontend/packages/portal-frontend/src/correlationAnalysis/hooks/useCorrelationAnalysisData.ts
```diff
@@ -77,8 +77,10 @@ export function useGeneCorrelationData(
 
         const filtered = rawCorrelates.filter(
           (cor: SortedCorrelations) =>
-            cor.feature !== featureName &&
-            cor.featureDatasetGivenId !== selectedDataset.datasetId
+            !(
+              cor.feature === featureName &&
+              cor.featureDatasetGivenId === selectedDataset.datasetId
+            )
```
rcreasi, line 83: I think I prefer your suggestion of 
```ts
        const filtered = rawCorrelates.filter(
          (cor: SortedCorrelations) =>
            cor.feature !== featureName ||
            cor.featureDatasetGivenId !== selectedDataset.datasetId
        );
```
I have to turn `!(A && B)` into `!A || !B` in my head before I can make intuitive sense of it. But that might just be me.

rcreasi: I grapple with "it must not be the case that both things are true" more than "either one could be false."

rcreasi (approved):

PR #712, opened 2026-07-01
Fix problem where SQL queries completely break under load.

pgm:
This is a successor to https://github.com/broadinstitute/depmap-portal/pull/709 because the problem continued after that PR. 

Based on the linked github issues people seemed to be reporting a set of issues that were stemming from multiple threads reading from the same redis connection. Now there's a connection pool that is supposed to give one connection to each thread so that situation is impossible.

However, from the analysis that claude did (described on that PR) it sounds like there might be an issue with doing a blocking wait on a task (which uses pubsub in redis, which the claude's comment suggest might be using a cached connection. If that's really true, then it's logical that would leads the connection to be used across threads which would result in the error we're seeing.

To avoid this, switched from a blocking `task.get()` to polling `task.ready()`

I was able to reproduce the issue with this bash script:

```
jitter() { sleep $(awk 'BEGIN{srand(); print rand()}'); }
for i in {1..100}; do
  echo "$i"
  sh sample-sql-query-3.sh &
  jitter
done
```

where `sample-sql-query-3.sh` is
```
curl -X 'POST' \
  'http://localhost:8018/depmap-istaging/breadbox/temp/sql/query' \
  -H "X-Forwarded-Email: noone" \
  -H 'accept: text/csv' \
  -H 'Content-Type: application/json' \
  -d '{
  "sql": "SELECT sample_id, value\nFROM hotspot_mutations_internal_26q\nWHERE feature_id = '\''596'\'' AND sample_id in ( '\''ACH-000697'\'','\''ACH-000157'\'')"
}
```

After making the change, the queries appear to work as expected.

### DIFF
--- a/breadbox/breadbox/api/temp/sql.py
+++ b/breadbox/breadbox/api/temp/sql.py
@@ -101,22 +101,23 @@ async def _run_time_bounded_celery_task(
     """
     task = task_fn.apply_async(args=args, time_limit=time_limit)
 
-    # Use propagate=False so task.get() returns normally instead of raising the task's
-    # exception. When propagate=True (the default), a task failure causes the exception
-    # to propagate through Celery's generator-based drain_events_until/_wait_for_pending
-    # chain. Python doesn't close generators when an exception escapes a for-loop, so
-    # their finally blocks (including the Redis pub/sub unsubscribe) never run. The shared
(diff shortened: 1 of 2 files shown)


snwessel (approved):
This looks great to me!

PR #741, opened 2026-08-21
fix(breadbox): Tabular upload memory usage

snwessel:
**Background**: We now have larger tabular datasets than ever (now up to 300,000 rows and 12 columns), which means we are now uploading millions of cells of data at a time. This created issues in our staging environment, where the upload is currently failing to run because of memory issues. 

**Debugging**: I ran the process locally with real data and did some profiling using `memray`. With our current implementation, the loading of the this metadata consumes ~6.6GB of memory, most of which can be traced back to the `add_tabular_dimensions` function in `breadbox/crud/dataset.py`. 

**The solution:** 
1. Loading and saving `TabularCells` one column at a time (and saving/flushing things to the DB after each column) 
2. Bulk inserting dictionary values instead of defining ORM instances of TabularCell. This dramatically cuts down on memory usage because we do still have ~300k TabularCells per column

_Performance before/after this change_: 
* Before: The background worker's process consumed ~6.6GB of memory
* After: The background worker's process only used ~0.5GB of memory

**Potential Future Concerns**: There may be other parts of our code which load all of the TabularCells into memory at once, which would likely also overload our staging environment. I will need to do some follow-up work to check whether that's the case.

### DIFF
--- a/breadbox/breadbox/crud/dataset.py
+++ b/breadbox/breadbox/crud/dataset.py
@@ -319,7 +319,6 @@ def add_tabular_dimensions(
     Adds tabular dataset dimensions to database.
     """
     dimensions = []
-    values = []
 
     for col in data_df.columns:
         annotation_id = str(uuid4())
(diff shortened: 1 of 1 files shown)


pgm (approved):
LGTM

PR #742, opened 2026-08-24
Update gene card URL

rcreasi:
Current: http://www.genecards.org/cgi-bin/carddisp.pl?gene=BRCA1
New: https://www.genecards.org/card/BRCA1


pgm (approved):
Thank you!

